PR #37, opened 2020-03-25
Add Firebase Analytics + change wording on sms/email/labels

Abibibi:


public/index.html
```diff
@@ -45,5 +45,14 @@
       To begin the development, run `npm start` or `yarn start`.
       To create a production bundle, use `npm run build` or `yarn build`.
     -->
+    <!-- The core Firebase JS SDK is always required and must be listed first -->
+    <script src="/__/firebase/7.12.0/firebase-app.js"></script>
```
HugoGresse: no need to import firebase script, already done via npm in src/firebase/firebase.js

HugoGresse: same for the two below.

Abibibi: removed them in last commit

Issue #76, opened 2020-03-30
[Front] Country dropdown on sms field


HugoGresse:
In order to know which country the user is in, we have to let him choose it. 
So a dropdown will be a good idea with country flag + country name maybe. 

the default value should be the navigator language (it's not always accurate but it should cover more use case than having France as default country)

Phone given to the function should either be : 
06XXXXXXXX (user input)
+336XXXXXXXXXX (user input)
+336XXXXXXXXXX (+33 given by the dropdown, 6XXXXXXX by the user)

PR #106, opened 2020-04-07
[FRONT] :sparkles: add country select for sms field and lang menu

ntchou:
https://github.com/Instant-Visio/InstantVisio-WebApp/issues/76


ntchou:
image

ntchou:
There is a lot of changes because i push flag svg icons (220)

public/locales/fr/countries.json
```diff
@@ -0,0 +1,224 @@
+ {   
+	 "entries": {
```
HugoGresse, line 2: is there no library for country listing with translation?

ntchou: I found a lib for country listing with translation https://github.com/michaelwittig/node-i18n-iso-countries
For os/browser emoticon you'r right i didn't think about it. It will be better. I'll implement it

ntchou: i tried with browser emoticon it doesn't work. I'll try third party lib like https://github.com/lipis/flag-icon-css

src/components/Form/PhoneField.js
```diff
@@ -0,0 +1,55 @@
...
+        memoizedSetValueFunc(lang)
+    }, [memoizedSetValueFunc])
+
+    const onCountrySelect = (value) => {
+        memoizedSetValueFunc(value)
+    }
+
+    return (<Wrapper>
+        <Field
+            prepend={lang && <Country defaultCountry={lang} onSelect={onCountrySelect} />}
```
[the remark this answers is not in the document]
ntchou: You're right. I did it because `setValue` from formik useField hook triggered an infinite rerender. I had to memoize the function. I fix it. I inject the default country (retrieved from browser lang)  to phone field component

src/components/App/index.js
```diff
@@ -34,7 +34,7 @@ const App = () => {
                 viewport.setAttribute('content', `height=${viewheight}px, width=${viewwidth}px, initial-scale=1.0`)
             }, 300)
         }
-    }, [])
+    }, [i18n])
```
HugoGresse, line 37: why do we need i18n on the useEffect? maybe just window.location.pathname?

src/components/Country/index.js
```diff
@@ -0,0 +1,90 @@
...
+    const countries = useMemo(() => {
+        let translatedCountries = isoCountries.getNames(language)
+
+        if (!Object.entries(translatedCountries).length){
+            translatedCountries = isoCountries.getNames(options.fallbackLng[0])
+        }
+
+        let countriesList = Object.entries(translatedCountries)
+        if (selectedCountries.length){
+            countriesList = countriesList.filter(([code]) => selectedCountries.includes(code))
```
HugoGresse, line 42: I think the filter could be done directly, without checking the length of the array, if the array exist.

src/components/Footer/index.js
```diff
@@ -55,4 +55,4 @@ const Footer = () => {
     )
 }
 
-export default Footer
+export default React.memo(Footer)
```
HugoGresse, line 58: why do we need a memo on the Footer?

src/pages/Home/Description.js
```diff
@@ -143,7 +141,7 @@ export default function Description(){
         <Information>
             <P {...(isMobile && {onClick:() => setCollapsed(!collapsed)})}>
                 <span>{t('information.title')}</span>
-                <Arrow collapsed={collapsed} />
+                <Arrow collapsed={collapsed}>&#x2BC6;</Arrow>
```
HugoGresse, line 144: I think we need a real icon here, the emoticion was not displayed on my side :/

src/components/App/index.js
ntchou: it's an oversight. I'll fix it

src/components/Country/index.js
ntchou: yes

src/pages/Home/Description.js
ntchou: For me it's ok. I'll change for an icon

src/components/Footer/index.js
ntchou: I saw when i choose a lang it was rerender 4 times

PR #110, opened 2020-04-12
brought changes based on @ntchou's review in feature/103-full-screen

Abibibi:


src/pages/VideoCall/index.js
```diff
@@ -50,35 +50,24 @@ const VideoCallFrame = () => {
...
-                    setParticipantNumber(4)
-                    break
-                default:
-                    setParticipantNumber(1)
+                if (participantsLength < 2) {
+                    setParticipantStatus(t('waiting-participant'))
+                } else if (participantsLength === 2 && (event && !event.participant.local)) {
+                    setParticipantStatus(!event.participant.video ? t('participant-cam-off') : '')
+                } else {
+                    setParticipantStatus('')
```
ntchou, line 67: Define `setParticipantStatus` default value with `(t('waiting-participant'))`

We can write it more simply. No need to check if participants is defined

```
setParticipantNumber(participantsLength)

if (participantsLength === 2 && (event && !event.participant.local)) {
    setParticipantStatus(!event.participant.video ? t('participant-cam-off') : '')
}
```

src/pages/VideoCall/index.js
```diff
@@ -94,17 +83,10 @@ const VideoCallFrame = () => {
...
-
-            if (event.participant.local === false && Object.keys(participants).length < 2) {
-                setParticipantStatus(t('participant-left'))
-            } else {
-                setParticipantStatus('')
-            }
-        })
+            .on('participant-updated', eventActions)
+            .on('participant-left', (event) => { 
+                setParticipantStatus(!event.participant.local && Object.keys(participants).length < 2 ? t('participant-left') : '')
```
ntchou, line 88: `participantsLength` is define above use it instead ` Object.keys(participants).length `

src/pages/VideoCall/index.js
Abibibi: Thanks, done. Still kept a `participantsLength > 2` check because participantStatus is displayed only when there are 2 participants (not more, knowing that calls can welcome up to 4 participants for now).

src/pages/VideoCall/index.js
Abibibi: `participantsLength` is defined in the `eventActions` method. However, `eventActions` is not called when `participant-left` event occurs, hence why `Object.keys(participants).length` is used. 

To avoid this and to be able to use `participantsLength` everywhere, I tried to:

- define `participantsLength` outside of the eventActions method
- put participantsLength in the state (as such: `const [ participantsLength, setParticipantsLength ] = useState(0) `)

But then, the participant status message is not displayed properly, so I kept the code as is.

ntchou: Ok

src/pages/VideoCall/index.js
ntchou: Ok

PR #120, opened 2020-04-14
:bug: fix lang detector. Use default option

ntchou:


HugoGresse:
There is an issue where on first launch, localStorage is empty, selected lang is "fr-FR" and the top right lang menu is defaulted to LANG.FR-FR
Capture d’écran 2020-04-14 à 17 05 10

ntchou:
Not same behavior on firefox. I'll fix it
image

ntchou:
I didn't see the issue because my browser configured with french lang and not french (france). Now i see the issue ;). I'll fix it

PR #126, opened 2020-04-15
Legal mention fixes + add exacte adresse 

HugoGresse:
- remove encoded unicode char
- remove unused legal-mentions.md 
- add exact adresse for Stéphane (04860 PIERREVERT, France, Terre, Système Solaire, Nuage interstellaire local, Bulle locale, Bras d'Orion, Voie lactée)
- fix malformed json file for Romanian legal mention


src/pages/VideoCall/VideoCall.js
```diff
@@ -4,7 +4,7 @@ import {SCREEN} from '../../styles/theme'
 
 export const CallContainer = styled.div`
     width: 100vw;
-    height: 100%;
+    height: 100vh;
```
Abibibi, line 7: height needs to be back to 100%, as well as in html and body, otherwise the app is not well displayed on Android Chrome.

HugoGresse: I've just checked and did not have any issue on android chrome.

HugoGresse: I've replaced height by min-height to allow the content to be bigger though (was not needed on my resolution though)

PR #136, opened 2020-04-16
new credit

ntchou:
image

### DIFF
--- a/src/pages/Credits/data.json
+++ b/src/pages/Credits/data.json
@@ -1,0 +1,78 @@
+{
+    "creators":[
+       {
+          "name":"Stéphan Luçon",
+          "link":"https://fr.linkedin.com/in/stephanelucon"
+       }
+    ],
+    "developers":[
+       {
+          "name":"Abeba Ngwe",
+          "link":"https://abebangwe.com"
+       },
+       {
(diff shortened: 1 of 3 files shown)


src/pages/Credits/data.json
```diff
@@ -0,0 +1,78 @@
+{
+    "creators":[
+       {
+          "name":"Stéphan Luçon",
+          "link":"https://fr.linkedin.com/in/stephanelucon"
+       }
+    ],
+    "developers":[
```
HugoGresse, line 8: two order way: 
randomize on each runtime
family name alphabetical order.

src/pages/Credits/data.json
```diff
@@ -0,0 +1,78 @@
...
+    ],
+    "partners":[
+       {
+          "name":"Daily.co",
+          "link":"https://www.daily.co3"
+       },
+       {
+          "name":"OVHcloud",
+          "link":"https://www.ovh.com"
+       }
```
HugoGresse, line 50: Twilio too

src/pages/Credits/data.json
ntchou: why do you want to randomize on each runtime?

PR #146, opened 2020-04-22
Feature/142 no translate green cam mic chat on video call

Abibibi:


src/pages/VideoCall/VideoCall.js
```diff
@@ -133,6 +133,10 @@ export const Controls = styled.div`
     .cam-audio {
         display: flex;
         flex-direction: row;
+
+        p {
```
HugoGresse, line 137: :D

PR #156, opened 2020-04-27
:sparkles: gdpr management

ntchou:


package.json
```diff
@@ -33,6 +33,7 @@
         "i18next-browser-languagedetector": "^4.0.2",
         "i18next-xhr-backend": "^3.2.2",
         "libphonenumber-js": "^1.7.49",
+        "lodash": "^4.17.15",
```
HugoGresse, line 36: lodash is very heavy, we could probably go without it without major issue don't you think?

ntchou: Yes it was for simplicity. I'll remove it. I wanted to use optional chaining but I don't know if it's already ready. I saw it didn't work on Firefox android

HugoGresse: maybe CRA pollyfil it?

src/utils/cookies.js
```diff
@@ -0,0 +1,27 @@
...
+const hasCookiebot = () => window.Cookiebot
+
+const cookieBotAcceptListener = () => {
+    if (hasCookiebot() && window.Cookiebot.consent?.statistics) {
+        firebase.analytics()
+    }
+}
+
+const cookieBotDeclineListener = () => {
+    if (hasCookiebot() && window.Cookiebot.consent?.statistics) {
```
HugoGresse, line 13: this is crazy /o/ 
what a day we leave in.

src/utils/cookies.js
```diff
@@ -0,0 +1,27 @@
+import firebase from 'firebase/app'
+import 'firebase/analytics'
+
+const hasCookiebot = () => window.Cookiebot
```
HugoGresse, line 4: Having Google anaytics logic within a cookie.js make it weird, I'll have either split that in two files (exporting it from the firebase.js for eg) or have an analytics.js with the cookie logic inside or something like this.

src/utils/cookies.js
ntchou: mistake

PR #160, opened 2020-04-29
Feature/159 credits css improvement delete text under home desc

Abibibi:


public/locales/it/home.json
```diff
@@ -9,11 +9,11 @@
         },
         "indications": {
             "landscape": "Dopo aver inviato il formulario, ti consigliamo di impostare il tuo dispositivo in modalità paesaggio. Potrai vedere meglio il tuo familiare.",
```
ntchou, line 11: if we don't use `landscape` translation anymore we can remove it

Abibibi: ok, done

PR #188, opened 2020-05-05
Enhancement/166 improved ux ui home mobile tablet

Abibibi:
To make sure form is visible as soon as possible when arriving on homepage.

Fold is now on both mobile and tablet:

- to make as much space as possible for form
- because UX-wise, it still makes more sense to put instructions before form

newTabletHomeUI
newPhoneHomeUI


src/hooks/useDetectMobile.js
```diff
@@ -11,5 +11,5 @@ export default function useDetectMobile() {
             window.removeEventListener('resize', handleResize)
         }
     }, [])
-    return width < 768
+    return width < 1180
```
HugoGresse, line 14: did you mean to edit the function for useDetectMobileOrTablet?

Abibibi: Yes ! it's now edited :)

PR #189, opened 2020-05-07
Minor improvement on call state

HugoGresse:


src/pages/VideoCall/index.js
```diff
@@ -17,8 +17,8 @@ import ErrorDialog from './ErrorDialog'
 const VideoCallPage = () => {
     const { t } = useTranslation('videocall')
 
-    const [camOn, setCamOn] = useState(true)
-    const [audioOn, setAudioOn] = useState(true)
+    const [camOn, setCamOn] = useState(false)
+    const [audioOn, setAudioOn] = useState(false)
```
Abibibi, line 21: Yes, it was previously set to false too. 

However, some testers said that it was confusing to them, that the video and audio icons initially appeared deactivated and then activated, once the video call fully loaded (that is in the case that cam and mic were previously authorized by the user).

HugoGresse:
thanks @Abibibi !

PR #198, opened 2020-05-12
Fix/197 mute camera div always displayed

Abibibi:


Abibibi:
> lgtm, maybe just remove the yarn.lock & add it to gitignore?

done @HugoGresse  :)

PR #211, opened 2020-05-26
In the News page done

Abibibi:


src/pages/MediaNews/index.js
```diff
@@ -0,0 +1,35 @@
...
+import data from './data'
+
+import DefaultLayout from '../../layout/Default'
+
+const MediaNews = () => {
+    const { t } = useTranslation('media')
+
+    return (
+        <DefaultLayout title={`${t('page-title')} - Instant Visio`}>
+            <MediaNewsTitle>On parle de nous</MediaNewsTitle>
```
ntchou, line 13: missing translation

public/locales/gr/media.json
```diff
@@ -0,0 +1,4 @@
+{
+    "page-title": "On parle de nous",
```
HugoGresse, line 2: Maybe worth using english for all non french / non translated lang?

src/pages/MediaNews/index.js
Abibibi: Done

public/locales/gr/media.json
Abibibi: Yes! Done

PR #266, opened 2020-10-25
Feature/ios icon and splashscreen

kangadominique123:
Simulator Screen Shot - iPhone X - 2020-10-25 at 17 15 36
Simulator Screen Shot - iPhone X - 2020-10-25 at 17 15 42
Simulator Screen Shot - iPhone X - 2020-10-25 at 17 15 47


MattiaPrimavera (approved):
Looks good to me, only the background icon is still the ionic default one. I'm opening another issue for that one, since it's less important than the main icon & splash screen

HugoGresse (approved):

PR #270, opened 2020-10-27
Fix video call issue on WebApp

MattiaPrimavera:
- fix video call issue in web app (error occurred after merging mobile PRs)
- add build/run instructions for mobile apps (`README.md`)
- add run npm scripts for mobile apps


MattiaPrimavera:
Build failed with:

```bash
./src/pages/VideoCall/index.js
89
  Line 151:8:  React Hook useEffect has a missing dependency: 'isMobile'. Either include it or remove the dependency array  react-hooks/exhaustive-deps
```

whereas it passes on my local (`npm run build`)

PR #276, opened 2020-11-08
API Bootstrap

HugoGresse:
- add new API express app
- authentificate those calls using an express middleware
- token are JWT, the first token for each user is generated upon login to users/uid/ [tokens]: tokenId: {createdAt, valid}
- swagger route to get the API specs
- a login button very ugly on the front
- fix for greek lang


functions/src/api/authentificateJWT.spec.ts
```diff
@@ -0,0 +1,5 @@
+describe('authentificateJWT', () => {
+    it('should test that true === true', () => {
+        expect(true).toBe(true)
```
HugoGresse, line 3: this will be done later on.

functions/src/api/authentificateJWT.spec.ts
```diff
@@ -0,0 +1,5 @@
+describe('authentificateJWT', () => {
```
MattiaPrimavera, line 1: spelling errror: authentificate -> authenticate, as it should follow the file name `authenticateJWT`

functions/src/api/authentificateJWT.ts
```diff
@@ -0,0 +1,55 @@
+import * as jsonWebToken from 'jsonwebtoken'
+import { Request, Response } from 'express'
+import * as functions from 'firebase-functions'
+import { VerifyErrors } from 'jsonwebtoken'
+import { JWTData } from '../types/JWTData'
+import { JWTToken } from '../types/JWTToken'
+import { isTokenValidInDb } from '../db/isTokenValidInDb'
+
+export const authenticateJWT = (
```
MattiaPrimavera, line 9: spelling error in the file name: `authentificateJWT.ts` -> `authenticateJWT.ts`

functions/src/api/authentificateJWT.ts
```diff
@@ -0,0 +1,55 @@
...
+            {
+                algorithms: ['HS256'],
+            },
+            async (err: VerifyErrors | null, data: {} | undefined) => {
+                if (err) {
+                    return res.sendStatus(403)
+                }
+                if (!data) {
+                    return res.sendStatus(403)
+                }
```
MattiaPrimavera, line 39: pls consider to squash the two if in a single one and use `OR` operator, since the response status is the same

ntchou:
Good job

functions/src/api/authentificateJWT.ts
```diff
@@ -0,0 +1,55 @@
...
+                const jwtData: JWTData = <JWTData>data
+
+                const isTokenValid = await isTokenValidInDb(jwtData, token)
+
+                if (!isTokenValid) {
+                    return res.sendStatus(403)
+                }
+
+                next()
+                return
```
ntchou, line 49: Useless

functions/src/db/isTokenValidInDb.ts
```diff
@@ -0,0 +1,23 @@
...
+): Promise<boolean> => {
+    const result = await db.collection('users').doc(jwtData.uid).get()
+
+    if (!result || !result.exists) {
+        return false
+    }
+
+    const userData = <UserData>result.data()
+
+    return (
```
ntchou, line 18: `return Boolean(.....)`

functions/src/api/authentificateJWT.ts
```diff
@@ -0,0 +1,55 @@
...
+                }
+                if (!data) {
+                    return res.sendStatus(403)
+                }
+                const jwtData: JWTData = <JWTData>data
+
+                const isTokenValid = await isTokenValidInDb(jwtData, token)
+
+                if (!isTokenValid) {
+                    return res.sendStatus(403)
```
MattiaPrimavera, line 45: although the status are well known, I'd prefer to read explicit names instead of numbers for the status codes:
`403` -> `statusCode.FORBIDDEN` ?

functions/src/api/swaggerDefinition.ts
```diff
@@ -0,0 +1,15 @@
+export const swaggerDefinition = {
+    openapi: '3.0.0',
+    info: {
+        version: '1.0.0',
+        title: 'InstantVisio API',
+        description:
+            'Provide an open source API to create, joins, manage video rooms, and more',
```
MattiaPrimavera, line 7: spelling error: `joins` -> `join`

functions/src/db/isTokenValidInDb.ts
```diff
@@ -0,0 +1,23 @@
...
+import { db } from '../firebase/firebase'
+import { UserData } from '../types/UserData'
+
+export const isTokenValidInDb = async (
+    jwtData: JWTData,
+    token: JWTToken
+): Promise<boolean> => {
+    const result = await db.collection('users').doc(jwtData.uid).get()
+
+    if (!result || !result.exists) {
```
ntchou, line 12: `if (!result?.exists)` if we're in node 14

functions/src/db/isTokenValidInDb.ts
```diff
@@ -0,0 +1,23 @@
+import { JWTData } from '../types/JWTData'
+import { JWTToken } from '../types/JWTToken'
+import { db } from '../firebase/firebase'
+import { UserData } from '../types/UserData'
+
+export const isTokenValidInDb = async (
+    jwtData: JWTData,
+    token: JWTToken
+): Promise<boolean> => {
+    const result = await db.collection('users').doc(jwtData.uid).get()
```
ntchou, line 10: Typing?

functions/src/db/addTokenToUser.ts
```diff
@@ -0,0 +1,33 @@
+import { UID } from '../types/uid'
+import { JWTToken } from '../types/JWTToken'
+import { db, serverTimestamp } from '../firebase/firebase'
+
+export const addTokenToUser = async (
+    userId: UID,
+    token: JWTToken
+): Promise<{ success: boolean }> => {
```
MattiaPrimavera, line 8: why returning an object instead of directly a `boolean` here ?

functions/src/triggers/userCreate.ts
```diff
@@ -0,0 +1,21 @@
...
+import { addTokenToUser } from '../db/addTokenToUser'
+
+export const userCreate = functions.auth.user().onCreate(async (user) => {
+    const { jwt } = functions.config()
+    if (!jwt.key) {
+        throw new functions.https.HttpsError(
+            'failed-precondition',
+            'Missing JWT Key'
+        )
+    }
```
MattiaPrimavera, line 12: this lines could be moved in a `utility` file: pls consider getting the token through a utility function:

`const {jwt} = getJwtToken()`

This would eliminate duplication in `authentificateJWT.ts`

src/firebase/firebase.tsx
```diff
@@ -12,8 +13,9 @@ const firebaseConfig = {
     measurementId: process.env.REACT_APP_MEASUREMENT_ID,
 }
 
-firebase.initializeApp(firebaseConfig)
-
+const firebaseInstance = firebase.initializeApp(firebaseConfig)
+export const firebaseAuth = firebase.auth
+export const authInstance = firebaseInstance.auth()
```
MattiaPrimavera, line 18: did you mean to use `firebaseAuth` variable created above here ?

MattiaPrimavera (changes requested):
Generally good! Just some spelling errors and few considerations on the code

functions/src/db/isTokenValidInDb.ts
HugoGresse: doesn't the ```

        userData.tokens &&
        userData.tokens[token] &&
        userData.tokens[token].valid
```
already produce a boolean?

File: functions/src/db/isTokenValidInDb.ts (hunk 9 above)
Comment (HugoGresse) [reply functions/src/db/isTokenValidInDb.ts:12]: TS 4.0 does it too. thanks.

File: functions/src/db/isTokenValidInDb.ts (hunk 10 above)
Comment (HugoGresse) [reply functions/src/db/isTokenValidInDb.ts:10]: it's infered with the tyeps from Firestore.

File: functions/src/api/authentificateJWT.ts (hunk 7 above)
Comment (HugoGresse) [reply functions/src/api/authentificateJWT.ts:45]: https://fr.wikipedia.org/wiki/Liste_des_codes_HTTP 
:D 
un peu de par coeur même si les deux son valide :D

File: functions/src/db/addTokenToUser.ts (hunk 11 above)
Comment (HugoGresse) [reply functions/src/db/addTokenToUser.ts:8]: it makes things more clear about what's expected for the return.

File: functions/src/api/authentificateJWT.ts (hunk 5 above)
Comment (HugoGresse) [reply functions/src/api/authentificateJWT.ts:49]: Error : not all code return a value TS7030

File: src/firebase/firebase.tsx (hunk 13 above)
Comment (HugoGresse) [reply src/firebase/firebase.tsx:18]: no, we could have many Firebase config within a single project, this specificly require the firebaseInstance auth rather than the global defined one. In our case, it result in the same behavior though.

**Comment by HugoGresse** [comment]
Review change pushed!

File: functions/src/db/isTokenValidInDb.ts (hunk 6 above)
Comment (ntchou) [reply functions/src/db/isTokenValidInDb.ts:18]: my bad i didn't check `UserData` type

File: functions/src/api/authentificateJWT.ts (hunk 7 above)
Comment (MattiaPrimavera) [reply functions/src/api/authentificateJWT.ts]: :P I know them as well :D just consider the explicit words more readable and elegant :) but it's just an opinion, it's ok to me if you prefer to keep the numbers

File: functions/src/db/addTokenToUser.ts (hunk 11 above)
Comment (MattiaPrimavera) [reply functions/src/db/addTokenToUser.ts:8]: not sure I agree on this one. Why not to:
- raise an `Error` when the operation fails
- do not return anything in case the operation succeeds

File: functions/src/api/authenticateJWT.ts [hunk 14]
```diff
@@ -0,0 +1,46 @@
...
+import { JWTToken } from '../types/JWTToken'
+import { isTokenValidInDb } from '../db/isTokenValidInDb'
+import { assertJWTEnv } from '../utils/assertConfig'
+
+export const authenticateJWT = (
+    req: Request,
+    res: Response,
+    next: Function
+) => {
+    const jwtKey = assertJWTEnv()
```
MattiaPrimavera, line 14: I would use `getJWTKey` here, since:
- you're expecting a value as a return
- for consistency: the variable you affect the result of the function to is called `jwtKey`

The `get` method can still raise an error if the instance is not found. But `assert` is normally used in tests: as a reader, I would not expect an `assert` to return a value

functions/src/api/authenticateJWT.ts
```diff
@@ -0,0 +1,46 @@
...
+                if (err || !data) {
+                    return res.sendStatus(403)
+                }
+                const jwtData: JWTData = <JWTData>data
+
+                const isTokenValid = await isTokenValidInDb(jwtData, token)
+
+                if (!isTokenValid) {
+                    return res.sendStatus(403)
+                }
```
MattiaPrimavera, line 37: please consider keeping a single failure exit point for this function:

```typescript
try {
   // when error occurs:
   throw new Error(...)
   // success case

} catch(e) {
   // error case
   return res.sendStatus(403)
}
```

this removes the duplication and increases readability

functions/src/api/v1/rooms/getRooms.ts
```diff
@@ -0,0 +1,20 @@
...
+ *     responses:
+ *       200:
+ *         description: All the rooms linked to this token
+ */
+export const getRooms = async (req: Request, res: Response) => {
+    try {
+        res.send('Hello world')
+    } catch (error) {
+        console.error(error) // eslint-disable-line no-console
+        res.status(500).end()
```
MattiaPrimavera, line 18: why not to use `res.sendStatus` for consistency here ?

MattiaPrimavera (changes requested):

functions/src/triggers/userCreate.ts
```diff
@@ -0,0 +1,16 @@
+import * as functions from 'firebase-functions'
+import * as jsonWebToken from 'jsonwebtoken'
+import { addTokenToUser } from '../db/addTokenToUser'
+import { assertJWTEnv } from '../utils/assertConfig'
+
+export const userCreate = functions.auth.user().onCreate(async (user) => {
+    const jwtKey = assertJWTEnv()
```
ntchou, line 7: same comment than Matthia https://github.com/Instant-Visio/InstantVisio-WebApp/pull/276/files#r519495823

functions/src/triggers/userCreate.ts
```diff
@@ -0,0 +1,16 @@
...
+
+    const newJWTToken = jsonWebToken.sign({ uid: user.uid }, jwtKey, {
```
ntchou, line 9: you can destructuring 

```suggestion
    const { uid } = user
    const newJWTToken = jsonWebToken.sign({ uid }, jwtKey, {
```

functions/src/triggers/userCreate.ts
```diff
@@ -0,0 +1,16 @@
...
+        algorithm: 'HS256',
+    })
+
+    await addTokenToUser(user.uid, newJWTToken)
```
ntchou, line 13: you can use previous desctructuring

functions/src/triggers/userCreate.ts
```diff
@@ -0,0 +1,16 @@
+import * as functions from 'firebase-functions'
+import * as jsonWebToken from 'jsonwebtoken'
+import { addTokenToUser } from '../db/addTokenToUser'
+import { assertJWTEnv } from '../utils/assertConfig'
+
+export const userCreate = functions.auth.user().onCreate(async (user) => {
```
ntchou, line 6:
```suggestion
export const userCreate = functions.auth.user().onCreate(async (user: JWTData) => {
```

functions/src/types/JWTToken.ts
```diff
@@ -0,0 +1 @@
+export type JWTToken = string
```
ntchou, line 1: Maybe regroup types `JWTKey` and `JWTToken` in one file `functions/src/types/JWT.ts`. 

```suggestion
export type JWTToken = string
export type JWTKey = string
```

src/components/Login.tsx
```diff
@@ -0,0 +1,44 @@
...
+                setIsLoggedIn(true)
+                console.log('user logged in', user)
+            } else {
+                // Logged out
+            }
+        })
+    })
+
+    return (
+        <div>
```
ntchou, line 34: use can use `<>` instead of `<div>`

src/components/Login.tsx
```diff
@@ -0,0 +1,44 @@
...
+            {!isLoggedIn && <StyledFirebaseAuth
+                uiConfig={uiConfig}
+                firebaseAuth={authInstance}
+            />}
+            {isLoggedIn && <Button onClick={() => { authInstance.signOut()}} >Sign out</Button>}
```
ntchou, line 39: missing translation

functions/src/db/addTokenToUser.ts
HugoGresse: my thinking is that throwing error are not convenient to manage with async/await, and having a clear idea of the result just simplify reading the code later on.

functions/src/api/authenticateJWT.ts
HugoGresse: changed! As a side not, android framework do use assert at runtime to check the input on a method.

functions/src/api/authenticateJWT.ts
HugoGresse: how would you suggest refactoringthe method?

functions/src/api/v1/rooms/getRooms.ts
HugoGresse: this is a dumb implementation, this will change when implementing it

functions/src/triggers/userCreate.ts
HugoGresse: I'm not sure one way is clearer than the other, and destructuring, while being efficient, also add more computing than direct access to the property.

HugoGresse: EDIT: agree if reused.

src/components/Login.tsx
HugoGresse: This is a dumb component that will be removed when integration is done.

src/components/Login.tsx
HugoGresse: This is a dumb component that will be removed when integration is done.

HugoGresse:
most changes are done, I need to work on the UT now!

functions/src/api/authenticateJWT.ts
MattiaPrimavera: could u pls provide an example link for my personal improvement ? :)

functions/src/db/addTokenToUser.ts
MattiaPrimavera: I see, but doesn't actually `async/await` allow to write a code like it was synchronous ? So what is the difference with `try/catch` error handling in asynchronous compared to synchronous code ?

My thinking is that it is easier to read error cases as error raised by the function, instead of returning a response no matter error/success case, than having to interpret the read and interpret the result

functions/src/triggers/userCreate.ts
MattiaPrimavera: if this function is a hook, I would say `onUserCreate`

MattiaPrimavera (changes requested):
I think the only "blocking" thing for me is [this](https://github.com/Instant-Visio/InstantVisio-WebApp/pull/276/files/08211f3e11eb8fa2c2b0b7595d38ae5dc8d100d3#r519454877). I believe we should choose a consistent way to treat similar cases in the code base.

functions/src/db/addTokenToUser.ts
HugoGresse: Two more arguments: 
1. You want to authentificate a user. They are two possible outputs: the user is authentificated, or he is not. Both are expected state, managed by the code. So a method like `authentificateUser(): boolean` could, if using Error, never return false because it will throw an Error if the auth did not work. And as JS/IDE are nore powerful enough, how would you check for the type of value returned by the function? Is a boolean an indication that the user is logged in or not? Having a strong typed return value make things more clear. And throwing an error mean than this was not expected, but it was, it is just a rejection, not an exception. 

2. How would you finely manage error using multiple async/await within the same method? So having a try/catch for each await is possible, but it's very verbosed, and in my opinion make code harder to read.

But on the other side, we could do something like this: https://www.smashingmagazine.com/2020/08/error-handling-nodejs-error-classes/#error-classes

functions/src/api/authenticateJWT.ts
HugoGresse: well, I believe to had read some in the past but was unable to find example now. My assumption is that it's code generated from annotation that produce this, like @NonNull .

functions/src/db/addTokenToUser.ts
MattiaPrimavera: My point of view on the arguments:
1. Indeed there's still the need for functions that return boolean values, and that's a good example. There are two possible states for a user in your example, none of which is an error, and the function correctly express that: `isAuthenticated` would either return `true` or `false`, according to the user's state. In the case I commented instead, it seems you're not verifying something but trying to perform an action: `addTokenToUser`. This action can either succeed or fail. If you try to perform an action, you need to know the action succeeded, in which case the normal flow would continue (return a value if some valuable information is returned by the function). You also need to know if the execution failed, in which case you can throw a specific error, and handle it in a catch block.

2. I would indeed use error-classes, as I proposed in the other comment. That allows to:
    - have a single error handling block for a function (that calls multiple async functions)
    - specify what kind of error occurred in a reusable and clear way
    - distinguish different kind of errors
    - have a single error exit point for the function

functions/src/api/authenticateJWT.ts
ntchou: even more 
```
async (err: VerifyErrors | null, data: {} | undefined) => {
  try {
    if (err || !data) {
      throw new ForbiddenError() // new type to define here
    }

    const jwtData: JWTData = <JWTData>data
    await assertTokenValidInDb(jwtData, token) // this can raise a ForbiddenError
    next()
  } catch (e) {
   next(e)
  }
}
```

and use an error middleware
```
errorMiddleware((error: Error, req, res, next) => {
  res.sendStatus(error.status)
})

app.use(errorMiddleware)
```

HugoGresse:
Ok, so I've implement real error and an Error middleware (thanks Nico!)
I did not went for the generic status code enum to be able to check the instance of the error. 
I also added the unit test where needed (not yet on the express as there is nothing yet). I did not use function-test module as the offline functionnality is crappy and does not help manage all cases, so I went with the overall mock of the firestore and functions. 
Let me know what you think

functions/src/api/auth/assertValidToken.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+import { ForbiddenError } from '../errors/HttpError'
+
+describe('assertValidToken', () => {
+    const token = 'tokenValue'
+    const jwtData: JWTData = {
+        uid: '69',
+        iat: 1544455,
+    }
+
+    it('should not assert because token is valid in DB', async () => {
```
ntchou, line 15:
```suggestion
    it('should assert because token is valid in DB', async () => {
```

functions/src/api/auth/assertValidToken.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+            exists: true,
+        }
+        firestoreGet.mockImplementation(() => Promise.resolve(snapshot))
+
+        const result = await assertValidToken(jwtData, token)
+
+        expect(result).toBeUndefined()
+    })
+
+    it('should not assert because token is valid in DB', async () => {
```
ntchou, line 33:
```suggestion
    it('should not assert because token is not valid in DB', async () => {
```

functions/src/api/auth/assertValidToken.spec.ts
```diff
@@ -0,0 +1,46 @@
+// tslint:disable-next-line
+import '../../testUtils/mockFirebaseAdminAndFunctions'
```
MattiaPrimavera, line 2: why the need to disable linting here ?

functions/src/api/auth/assertValidToken.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+                    },
+                },
+            }),
+            exists: true,
+        }
+        firestoreGet.mockImplementation(() => Promise.resolve(snapshot))
+
+        const result = await assertValidToken(jwtData, token)
+
+        expect(result).toBeUndefined()
```
MattiaPrimavera, line 30: please consider putting a try/catch block here as below. The catch block could have a failing expect (so that if the program reaches the catch, the test will fail), whereas the try block could have a passing expect, proving no exception was raised

functions/src/api/auth/assertValidToken.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+import { ForbiddenError } from '../errors/HttpError'
+
+describe('assertValidToken', () => {
+    const token = 'tokenValue'
+    const jwtData: JWTData = {
+        uid: '69',
+        iat: 1544455,
+    }
+
+    it('should not throw exception because token is valid in DB', async () => {
```
MattiaPrimavera, line 15: I would suggest to use affirmative sentences to describe tests. What about something like

```
it should confirm token validity in DB
```

functions/src/api/auth/assertValidToken.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+    it('should throw exception because token is NOT valid in DB', async () => {
+        const snapshot = {
+            data: () => ({}),
+            exists: true,
+        }
+        firestoreGet.mockImplementation(() => Promise.resolve(snapshot))
+
+        try {
+            await assertValidToken(jwtData, token)
+        } catch (error) {
```
MattiaPrimavera, line 42: if the code presents an error, and the `assertValidToken` function does not correctly raise an exception, wouldn't the test pass with a false positive ? (I think we miss a failing `expect` after the `await assert...`)

functions/src/api/middlewares/authenticateJWTMiddleware.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+    PreconditionFailedError,
+    UnauthorizedError,
+} from '../errors/HttpError'
+
+describe('authentificateJWTMiddleware', () => {
+    const mockNext = jest.fn()
+    const mockRes = jest.fn()
+
+    afterEach(() => {
+        mockNext.mockClear()
```
MattiaPrimavera, line 17: if mocks are cleared in `afterEach`, shouldn't they be created in a `beforeEach` callback ?

functions/src/api/middlewares/authenticateJWTMiddleware.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+        mockRes.mockClear()
+    })
+
+    it('should not be authorized due to missing authorization header', () => {
```
MattiaPrimavera, line 21: I suggest to use consistent test descriptions. I would go with the affirmative sentence like below:

```
should fail due to missing authorization header
```

functions/src/api/middlewares/authenticateJWTMiddleware.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+        expect(() => {
+            authenticateJWTMiddleware(
+                {
+                    headers: {},
+                } as any,
+                mockRes as any,
+                mockNext
+            )
+        }).toThrowError(UnauthorizedError)
```
MattiaPrimavera, line 30: I'm not sure this works. Aren't you passing a function to `expect` without executing it ? Please consider using a try/catch block here, like in other cases.

functions/src/api/middlewares/authenticateJWTMiddleware.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+                    },
+                },
+            }),
+            exists: true,
+        }
+        firestoreGet.mockImplementation(() => Promise.resolve(snapshot))
+
+        const mockNextWithDone = (params: any) => {
+            expect(params.message).not.toBe('Unable to verify the JWT Token')
+            done()
```
MattiaPrimavera, line 82: Please consider using `async/await` in tests (for consistency and readability) instead of `done` callbacks

functions/src/api/middlewares/authenticateJWTMiddleware.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    const token = <JWTToken>splittedHeader[1]
+
+    jsonWebToken.verify(
+        token,
+        jwtKey,
+        {
+            algorithms: ['HS256'],
+        },
+        async (err: jsonWebToken.VerifyErrors | null, data: {} | undefined) => {
```
MattiaPrimavera, line 39: pls consider extracting this function from `authenticateJWTMiddleware`

functions/src/db/isTokenValidInDb.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+    it('should return false if nothing in db', async () => {
+        const snapshot = {
+            data: () => ({}),
+            exists: true,
+        }
+        firestoreGet.mockImplementation(() => Promise.resolve(snapshot))
+
+        const result = await isTokenValidInDb(jwtData, token)
+        expect(result).toBeFalsy()
+    })
```
MattiaPrimavera, line 40: isn't this test case a more generic expression for `should return false if token is not existing in db` test ? I think this may be removed

functions/src/db/isTokenValidInDb.ts
```diff
@@ -0,0 +1,22 @@
+import { db } from '../firebase/firebase'
+import { UserData } from '../types/UserData'
+import { JWTData, JWTToken } from '../types/JWT'
+
+export const isTokenValidInDb = async (
+    jwtData: JWTData,
+    token: JWTToken
+): Promise<boolean> => {
+    const result = await db.collection('users').doc(jwtData.uid).get()
```
MattiaPrimavera, line 9: pls consider being using a more specific variable name: isn't `result` a `userDocRef` ?

MattiaPrimavera (changes requested):

functions/src/api/auth/assertValidToken.spec.ts
HugoGresse: side effect related to TSLINT because of the generic import

functions/src/api/auth/assertValidToken.spec.ts
HugoGresse: the test will fail if the exception is raised anyway.

functions/src/api/auth/assertValidToken.spec.ts
HugoGresse: this is true for isTokenValid, but not for the assertValidToken, the meaning is different and it makes sens based on the method behavior.

functions/src/api/auth/assertValidToken.spec.ts
HugoGresse: if it's the case, the error will not be an instance of ForbiddenError as it's specified manually from the code itself.

functions/src/api/middlewares/authenticateJWTMiddleware.spec.ts
HugoGresse: they are created once for all the test file, just the count on the mock are reset after each test to prevent polution.

functions/src/api/middlewares/authenticateJWTMiddleware.spec.ts
HugoGresse: this is not possible due to the middleware signature which need to call next() wherever it wants, the middleware does not return a promise.

functions/src/api/middlewares/authenticateJWTMiddleware.ts
HugoGresse: not agree, the code is small enough and explicit enough.

functions/src/api/middlewares/authenticateJWTMiddleware.spec.ts
HugoGresse: `authenticateJWTMiddleware` function is called inside the arrow function passed to the `expect`. It makes more sense for me because you can read the test more easily as a phrase. I would have use it on other test if it was possible too.

functions/src/db/isTokenValidInDb.spec.ts
HugoGresse: by bad, exists should have been set to false here.

HugoGresse:
last wording change pushed and more feedback for Mattia's review.

functions/src/db/isTokenValidInDb.ts
HugoGresse: I've changed it through it duplicate the data as the type is already a DocumentSnapshot.

functions/src/api/auth/assertValidToken.spec.ts
MattiaPrimavera: would the lint error persist if writing the import as `import * as mockFirebaseAdminAndFunctions from ...` ?

functions/src/api/auth/assertValidToken.spec.ts
MattiaPrimavera: That's nice. Although I still feel it is strange to test the return value for a function that does not return anything :)
What if we just return the promise (`return assertValidToken(...)`) like in [this post](https://stackoverflow.com/questions/54525147/expect-a-jest-test-to-resolve-but-dont-care-about-the-value) ?

functions/src/api/auth/assertValidToken.spec.ts
MattiaPrimavera: What I meant is: we are testing that the function raises an exception. What if the function does not instead ?

```typescript
try {
  await ...
  // what if the code execution goes here, since the function did not raise an exception as expected ? Will the test fail or pass ?
} catch(e) {
   expect(...) // We test the function raises an exception here, as expected
}
```

functions/src/api/middlewares/authenticateJWTMiddleware.ts
MattiaPrimavera: let's leave it this way in this case. Function length is still a little subjective, but if I correctly recall, Uncle Bob describes clean code functions as having an optimal length of like 4 lines. I wouldn't be too strict on the subject either, but I thought that declaring it as a separate function would have reduced the vertical length (a measure for complexity) of the function (which is like around 40 lines now).

HugoGresse: thanks for the feedback, I'll keep in mind for the next PR!

functions/src/api/auth/assertValidToken.spec.ts
HugoGresse: the test will fail because the expect from the catch was not called.

functions/src/api/auth/assertValidToken.spec.ts
HugoGresse: another one by importing an unused module in the file, TS is very picky!

MattiaPrimavera (approved):

ntchou (approved):

PR #282, opened 2020-11-16
Add redux & redux-thunk

MattiaPrimavera:
this includes configuration for using [Redux DevTools](https://github.com/zalmoxisus/redux-devtools-extension) (on chrome)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17811,6 +17811,28 @@
                 "warning": "^4.0.3"
             }
         },
+        "react-redux": {
+            "version": "7.2.2",
+            "resolved": "https://registry.npmjs.org/react-redux/-/react-redux-7.2.2.tgz",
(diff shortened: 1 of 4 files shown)


HugoGresse:
LGTM, maybe we can add a part to better integrate redux with the hot reload? 

ex: 

```



const store = configureStore({})

const render = AppComponent => {
    return ReactDOM.render(
        <Provider store={store}>
            <AppComponent />
        </Provider>,
        document.getElementById('root')
    )
}
render(App)

if (module.hot) {
    module.hot.accept('./App', () => {
        const NextApp = require('./App').default
        render(NextApp)
    })
}
```

MattiaPrimavera:
I see what you mean. I just made a test on my local env although, and without adding the config, if I add a little text in the page, the page hot-reloads, but the store is keeping the state. Is hot-reloading already keeping the state by default ? Strange although, because the official docs have a section on [hot-reloading](https://redux.js.org/recipes/configuring-your-store)

HugoGresse (approved):
regarding hot reload, we can change that later on if the behavior can be improved. 
LGTM

ntchou (approved):

PR #284, opened 2020-11-18
Rename component files

### LINKED ISSUE #283: Component files rename
### Problem
Currently components main file are called `index.tsx` and are located under a directory with the name of the component, like following:

```
./src/components/Arrow/index.tsx
```

this makes it harder to walk through the tree structure with the IDE shortcuts, since a lot of files are called `index.tsx`.

### Proposed solution
rename the component `index.tsx` file to the component name, as the directory one:

```
./src/components/Arrow/Arrow.tsx
```

MattiaPrimavera:
use files named by the component instead of
index.tsx, which makes searching through the code
base for components harder

fix #283


HugoGresse (approved):
blind approve!

ntchou (approved):
Same, blind approve

PR #285, opened 2020-11-21
API : Implement room creation

HugoGresse:
on request: 
- create a new Room in IV db, with a 6 digit password or the provided one
- create the equivalent room on twilio, which will last for 5 minutes (and will be recreated upon user join but that's another PR)
- update the room in db with the twilio room sid
- return room id and room sid


functions/src/api/v1/rooms/createRoom.ts
```diff
@@ -0,0 +1,61 @@
...
+ *               }
+ *       401:
+ *         description: missing authorization bearer token
+ *       403:
+ *         description: authorization header present but not valid
+ *       412:
+ *         description: authorization header present but not formatted correctly
+ */
+export const createRoom = async (req: Request, res: Response) => {
+    const roomPassword = req.body.password
```
ntchou, line 42:
```suggestion
    const roomPassword = req.body.password || ~~(Math.random() * 999999)
```

functions/src/types/Room.ts
```diff
@@ -0,0 +1,10 @@
+export interface Room {
+    roomId: RoomId
+    roomSid: RoomSid
+    parameters: RoomParameters
+}
+
+export type RoomId = string
+export type RoomSid = string
```
ntchou, line 8: declare them before `Room`

functions/src/types/Room.ts
```diff
@@ -0,0 +1,10 @@
...
+
+export interface RoomParameters {}
```
ntchou, line 10: declare them before `Room`

functions/src/api/v1/rooms/createRoom.ts
```diff
@@ -0,0 +1,61 @@
...
+        roomSid,
+        parameters: {},
+    })
+    res.send({
+        roomId,
+        roomSid,
+    } as NewRoomResponse)
+}
+
+export interface NewRoomResponse {
```
ntchou, line 58: declare it before `createRoom`

functions/src/api/v1/rooms/createRoom.ts
```diff
@@ -0,0 +1,61 @@
...
+export const createRoom = async (req: Request, res: Response) => {
+    const roomPassword = req.body.password
+        ? req.body.password
+        : Math.floor(Math.random() * 999999)
+    const roomId = await addRoom(res.locals.uid, roomPassword)
+    const roomSid = await createTwilioRoom(roomId)
+    await updateRoom({
+        roomId,
+        roomSid,
+        parameters: {},
```
ntchou, line 50: I think it could be omit

functions/src/types/Room.ts
```diff
@@ -0,0 +1,10 @@
+export interface Room {
+    roomId: RoomId
+    roomSid: RoomSid
+    parameters: RoomParameters
```
ntchou, line 4: see https://github.com/Instant-Visio/InstantVisio-WebApp/pull/285/files#r528222580. 
`paraleters?: RoomParameters`

HugoGresse:
Explamation mark record to beat 
Capture d’écran 2020-11-22 à 12 14 15

functions/src/api/v1/rooms/createRoom.ts
```diff
@@ -0,0 +1,61 @@
...
+ *       401:
+ *         description: missing authorization bearer token
+ *       403:
+ *         description: authorization header present but not valid
+ *       412:
+ *         description: authorization header present but not formatted correctly
+ */
+export const createRoom = async (req: Request, res: Response) => {
+    const roomPassword =
+        req.body.password || ~~(Math.random() * 999999)
```
ntchou, line 48: No need ternary operator. Only this
```
const roomPassword = req.body.password || ~~(Math.random() * 999999)
```
I don't know if github support multiline suggestions

ntchou (commented):
It's ok to me

functions/src/api/v1/rooms/createRoom.ts
```diff
@@ -0,0 +1,60 @@
+import { Request, Response } from 'express'
+import { addRoom } from '../../../db/addRoom'
+import { createTwilioRoom } from './service/createTwilioRoom'
+import { updateRoom } from '../../../db/updateRoom'
+import { RoomId, RoomSid } from '../../../types/Room'
+
+export interface NewRoomResponse {
+    roomId: RoomId
+    roomSid: RoomSid
```
MattiaPrimavera, line 9: what does the `roomSid` stand for ?

functions/src/api/v1/rooms/createRoom.ts
```diff
@@ -0,0 +1,60 @@
...
+    )
+    const roomSid = await createTwilioRoom(roomId)
+    await updateRoom({
+        roomId,
+        roomSid,
+    })
+    res.send({
+        roomId,
+        roomSid,
+    } as NewRoomResponse)
```
MattiaPrimavera, line 59: can't the object be created after line `51`, so that it can be used as a parameter in both calls to `.send` and `.updateRoom` ?

functions/src/types/Room.ts
```diff
@@ -0,0 +1,10 @@
+export type RoomId = string
+export type RoomSid = string
+
+export interface RoomParameters {}
+
+export interface Room {
+    roomId: RoomId
+    roomSid: RoomSid
```
MattiaPrimavera, line 8: may I ask why to type differently every string, to be more specific on what it means, considering the variable name will already hold that information ? I have the impression that, as a developer who reads it, I will understand the kind of information from the variable name already, but I'll have to go look into the `RoomId` type, to understand it is a simple `string`.

Also I find that the names are repeated. Why not to simplify with:

```typescript
export interface Room {
  id: string
  sid: string
}
```

so that when we'll access that info, it will be as `room.id` instead of `room.roomId`. In case you want to get the same name from destructuring, it will still be possible:

```typescript
const {id: roomId} = room
```

MattiaPrimavera (commented):
I left a few feedbacks. Good job!

functions/src/api/v1/rooms/createRoom.ts
HugoGresse: this is the twilio room sid that will maybe be needed later on

MattiaPrimavera: The difference being ? The `id` is the one we create, the `Sid` is the one created by the provider (twilio in this case) ?

HugoGresse: yes exactly

functions/src/types/Room.ts
HugoGresse: ensuring that a given parameter is of a specific type ensure that a given id is NOT wrongfully passed to a given parameter. 
Eg: RoomId will not be used as a RoomSid. A pontential counter argument will be that Typescript does not check for that specific use case, so I believe in this case we could use type Branding : https://stackoverflow.com/a/49432424/1377145

If we go for type branding (we probably should), indeed the class prefix is not needed.

MattiaPrimavera: Interesting, I'll read a little more on the subject

PR #286, opened 2020-11-22
API Room Edit

HugoGresse:
to be reviewed after #285 
- new patch room to edit room
- catch async error which is not possible within an express middleware.


functions/src/api/v1/rooms/createRoom.ts
```diff
@@ -57,4 +58,4 @@ export const createRoom = async (req: Request, res: Response) => {
         roomId,
         roomSid,
     } as NewRoomResponse)
```
ntchou, line 60: i think it's not usefull to precise type

HugoGresse: I remember why, it's to clarify for the client part which interface/response to expect.

ntchou: ok

maybe would be better ?
```
async (req: Request, res: Response):Promise<NewRoomResponse> => {
...
 res.send({
        roomId,
        roomSid,
})
}
``

HugoGresse: I can't as I'm not returning but calling res.send function, so probably extend or type the Response if there is a geneneric

ntchou: You're right. I review on my phone :)

PR #287, opened 2020-11-22
API room join

HugoGresse:
joining a room consist of:
- verifying the user as access to the API, token valid
- verifying the password match the given room id
- creating a JWT access token for the JS Client Twilio SDK.


functions/src/api/v1/rooms/service/createTwilioClientToken.ts
```diff
@@ -0,0 +1,23 @@
+import { Room } from '../../../../types/Room'
+import { UID } from '../../../../types/uid'
+import * as AccessToken from 'twilio/lib/jwt/AccessToken'
+import { getTwilioEnv } from '../../../../firebase/env'
+
+const VideoGrant = AccessToken.VideoGrant
+export const TTL_ACCESS_TOKEN_PARTICIPANT_SECONDS = 1440 // 24 hours
```
MattiaPrimavera, line 7: The [docs](https://www.twilio.com/docs/video/tutorials/user-identity-access-tokens#time-to-live) suggest:

> Time-To-Live (ttl)
> Access Tokens must be valid while joining a Room and when reconnecting to a Room due to network disruption or handoff. Therefore, we recommend that you set the ttl to the maximum allowed session length, which is currently 14,400 seconds (4 hours).

functions/src/db/getRoom.ts
```diff
@@ -0,0 +1,13 @@
...
+import { NotFoundError } from '../api/errors/HttpError'
+
+export const getRoom = async (roomId: RoomId): Promise<Room> => {
+    const documentSnapshot = await db.collection('rooms').doc(roomId).get()
+
+    if (!documentSnapshot.exists) {
+        throw new NotFoundError('Resource does not exist')
+    }
+
+    return <Room>documentSnapshot.data()
```
MattiaPrimavera, line 12: is type cast necessary here ? The function signature already specifies the return type

functions/src/firebase/env.ts
```diff
@@ -13,11 +13,20 @@ export const getJWTEnv = (): JWTKey => {
...
+    ) {
+        throw new InternalServerError(
+            'Missing Twilio sid or authtoken or apikeysid or apikeysecret'
+        )
     }
     return {
         sid: twilio.sid,
         authToken: twilio.authtoken,
+        apiKeySid: twilio.apikeysid,
+        apiKeySecret: twilio.apikeysecret,
```
MattiaPrimavera, line 30: why the case lack of consistency ?

MattiaPrimavera (changes requested):

functions/src/db/getRoom.ts
HugoGresse: indeed, but it is necessary, typescript doesn't get it.

functions/src/firebase/env.ts
HugoGresse: all env are lower case in instantvisio, I've followed this rules since the beginning, I don't have any clue why I choosed this but it's kind of impact full if we change this.

MattiaPrimavera (approved):

Issue #289, opened 2020-11-22
Use absolute path instead relative path


ntchou:
For a better re usability and maintainability. It's better to manage with absolute import

Issue #288, opened 2020-11-22
Create PremiumVideoCall page


MattiaPrimavera:
Inject twilio video react sample app code from [here](https://github.com/twilio/twilio-video-app-react)

### Remarks
- [ ] respect the `LICENSE`
- [ ] link the app to our main router
- [ ] check if env variables needed
- [ ] import the sample app tests

PR #290, opened 2020-11-22
Feat/front 289 absolute import

ntchou:
#289


HugoGresse:
I'm not sure I agree with that. Auto import will not work with absolute import and each import will need to be changed manually after :/ at least on Intellij Idea. 
What's the benefit?

ntchou:
It works if you set "typescript.preferences.importModuleSpecifier" to "non-relative". 
autoimport


And for the benefits:
- No more slashes and dots to reference a path
- You can move a file without having to update its input paths
- You can copy/paste code including imports into other files and not have to fiddle with the import paths.
- It's neat

HugoGresse:
>     * No more slashes and dots to reference a path
OK

>     * You can move a file without having to update its input paths
Already possible, at least in Intellij with the refactor tools 

>     * You can copy/paste code including imports into other files and not have to fiddle with the import paths.
Already the case, at least in Intellij, you can copy any code and the dependency will be added to the file on past. 

I hope the typescript settings will work will Intellij!

MattiaPrimavera:
> You can move a file without having to update its input paths

You should already be able doing this in VS code as well

> You can copy/paste code including imports into other files and not have to fiddle with the import paths.

Sure this does not work already in VS code instead ?

Are existing paths updated automatically ? I'm working on [this](https://github.com/Instant-Visio/InstantVisio-WebApp/issues/288): will all relative paths I inserted (a lot ...) be not valid anymore and have to be fixed manually ?

ntchou:
> > You can move a file without having to update its input paths
> 
> You should already be able doing this in VS code as well

Yes you're right

> > You can copy/paste code including imports into other files and not have to fiddle with the import paths.
> 
> Sure this does not work already in VS code instead ?

Yes it already does

> Are existing paths updated automatically ? I'm working on [this](https://github.com/Instant-Visio/InstantVisio-WebApp/issues/288): will all relative paths I inserted (a lot ...) be not valid anymore and have to be fixed manually ?

I don't think so

ntchou:
> > ```
> > * No more slashes and dots to reference a path
> > ```
> 
> OK
> 
> > ```
> > * You can move a file without having to update its input paths
> > ```

I was talking about vs code. I don't use Intellij. So i don't khow how it works.

> Already possible, at least in Intellij with the refactor tools
> 
> > ```
> > * You can copy/paste code including imports into other files and not have to fiddle with the import paths.
> > ```
> 
> Already the case, at least in Intellij, you can copy any code and the dependency will be added to the file on past.
> 
> I hope the typescript settings will work will Intellij!

PR #292, opened 2020-11-26
Implement getToken callable

HugoGresse:
Provide a callable function to get a valid token from the client side

### DIFF
--- a/functions/src/db/getUserToken.ts
+++ b/functions/src/db/getUserToken.ts
@@ -1,0 +1,26 @@
+import { UID } from '../types/uid'
+import { JWTToken } from '../types/JWT'
+import { db } from '../firebase/firebase'
+import { UserData } from '../types/UserData'
+
+export const getUserToken = async (userId: UID): Promise<JWTToken | null> => {
+    const userDocumentSnapshot = await db.collection('users').doc(userId).get()
+
+    if (!userDocumentSnapshot?.exists) {
+        return null
+    }
+
+    const userData = <UserData>userDocumentSnapshot.data()
+
+    if (userData.tokens) {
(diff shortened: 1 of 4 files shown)


functions/src/db/getUserToken.ts
```diff
@@ -0,0 +1,26 @@
+import { UID } from '../types/uid'
+import { JWTToken } from '../types/JWT'
+import { db } from '../firebase/firebase'
+import { UserData } from '../types/UserData'
+
+export const getUserToken = async (userId: UID): Promise<JWTToken | null> => {
+    const userDocumentSnapshot = await db.collection('users').doc(userId).get()
+
+    if (!userDocumentSnapshot?.exists) {
+        return null
```
MattiaPrimavera, line 10: for every created user, a token should normally exist. If it does not, I would say it is an error case

functions/src/db/getUserToken.ts
```diff
@@ -0,0 +1,26 @@
...
+        const validTokens = Object.keys(userData.tokens).filter(
+            (token) => userData.tokens[token].valid
+        )
+
+        if (validTokens.length > 0) {
+            return validTokens[0]
+        }
+    }
+
+    return null
```
MattiaPrimavera, line 25: this as well seems to me a *No valid token found* error

MattiaPrimavera: if this correctly returns `null`, then `getToken` as well would send back `null` to the client side, I guess

MattiaPrimavera (changes requested):

HugoGresse:
I need this for another PR, comments/review here will still be managed but in another PR.

PR #294, opened 2020-11-28
API: add invite participants route

HugoGresse:
- refactor newCall (free IV version) to move out the notification part. I didn't rewrited it fully, this will be done in another PR
- add new sendNotifications methods which manage both sms & email using a content and a destination
- add new route
- MISSING: save the number of email and sms sent to the user DB. 

I would appreciate any early feedback before finishing the stats part.


functions/src/api/v1/invite/inviteParticipants.ts
```diff
@@ -0,0 +1,144 @@
...
+            const country = dest.country || 'fr'
+            return {
+                ...dest,
+                lang,
+                country,
+            }
+        }
+    )
+
+    // TODO :
```
HugoGresse, line 111: this will be removed, see PR description

functions/src/notifications/sendEmail.ts
```diff
@@ -0,0 +1,41 @@
+import { NotificationParams } from '../types/Notification'
+import * as sgMail from '@sendgrid/mail'
+import { logEmailSent } from '../sumologic/sumologic'
+
+export const sendEmail = async (
```
HugoGresse, line 5: this is old IV code, you may not want to review it

functions/src/notifications/sendSMS.ts
```diff
@@ -0,0 +1,79 @@
...
+//             {'method': 'DELETE', 'path': '/sms/*/outgoing/*'}
+//         ]
+//     }).then((credential: any) => {
+//         console.log('auth success, open the url below to validate them', credential)
+//     }).catch((error: any) => {
+//         console.log('auth error', error)
+//     })
+// }
+
+export const sendSms = async (
```
HugoGresse, line 26: this is old IV code, you may not want to review it

functions/src/api/v1/invite/inviteParticipants.ts
```diff
@@ -0,0 +1,144 @@
...
+ *                           - de
+ *                           - es
+ *                           - gr
+ *                           - hu
+ *                           - it
+ *                           - ro
+ *
+ *     responses:
+ *       200:
+ *         description: One or many invites were sent successfully. It will return a 2xx if some invitation where not able to be sent (malformed email or phone number for eg).
```
ntchou, line 63: typo `It will return a 4xx`

ntchou (commented):
Tthis is my first feedbacks. I have others. It's a nice work

functions/src/firebase/env.ts
```diff
@@ -30,3 +33,42 @@ export const getTwilioEnv = (): TwilioEnv => {
...
+    if (!ovh.consumerkey || !ovh.servicename || !ovh.appsecret || !ovh.appkey) {
+        throw new InternalServerError(
+            'Missing ovh consumerkey or servicename or appsecret or appkey env'
+        )
+    }
+    return {
+        consumerkey: ovh.consumerkey,
+        servicename: ovh.servicename,
+        appsecret: ovh.appsecret,
+        appkey: ovh.appkey,
```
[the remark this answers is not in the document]
HugoGresse: as sayed in another PR, this are env declared on the servers, they where all lowercase since the beginning of the project and impact the production as well as the dev.

functions/src/api/v1/invite/inviteParticipants.ts
```diff
@@ -0,0 +1,144 @@
...
+    const notificationContent: NotificationContent = {
+        name: hostName,
+        roomUrl: roomUrl,
+    }
+
+    const notificationSentResult = await sendNotifications(
+        invitationsDestinations,
+        notificationContent,
+        ovhEnv,
+        sendGridEnv
```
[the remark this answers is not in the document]
HugoGresse: it seems to be better to to rely on env in those function and only execute it.

functions/src/api/v1/invite/inviteParticipants.ts
HugoGresse: no, this was legit, if the user want to send 5 emails and only 1 is sent, it will return a 2xx

functions/src/firebase/env.ts
```diff
@@ -30,3 +33,42 @@ export const getTwilioEnv = (): TwilioEnv => {
...
+export const getAppEnv = (): AppEnv => {
+    const { app } = functions.config()
+    if (!app.domain || !app.emailfrom) {
+        throw new InternalServerError('Missing app domain or emailfrom env')
+    }
+    return {
+        domain: app.domain,
+        emailFrom: app.emailfrom,
+    }
+}
```
MattiaPrimavera, line 46: pls consider excusing me for my love for destructuring :)

```js
export const getAppEnv = (): AppEnv => {
     const { app } = functions.config()
     const {domain, emailfrom} = app
     if (!domain || !emailfrom) {
         throw new InternalServerError('Missing app domain or emailfrom env')
     }
     return {
         domain
         emailFrom
     }
 }
```

functions/src/notifications/sendEmail.ts
```diff
@@ -0,0 +1,41 @@
...
+        text: messageBody,
+        ip_pool_name: params.sendGridCredentials.ip_pool_name,
+    }
+
+    try {
+        const result = await sgMail.send(msg)
+        const response = result && result[0]
+        if (response.statusCode > 200 && response.statusCode < 400) {
+            logEmailSent()
+            return Promise.resolve()
```
MattiaPrimavera, line 28: a simple `return` should be enough here

functions/src/notifications/sendEmail.ts
```diff
@@ -0,0 +1,41 @@
...
+        from: `InstantVisio <${params.emailFrom}>`,
+        subject: subject,
+        text: messageBody,
+        ip_pool_name: params.sendGridCredentials.ip_pool_name,
+    }
+
+    try {
+        const result = await sgMail.send(msg)
+        const response = result && result[0]
+        if (response.statusCode > 200 && response.statusCode < 400) {
```
MattiaPrimavera, line 26: same as above on conditions

```js
if(isEmailSendSuccessful()) {}
```

functions/src/notifications/sendEmail.ts
```diff
@@ -0,0 +1,41 @@
...
+import * as sgMail from '@sendgrid/mail'
+import { logEmailSent } from '../sumologic/sumologic'
+
+export const sendEmail = async (
+    params: NotificationParams,
+    messageBody: string,
+    subject: string
+) => {
+    if (!params.sendGridCredentials) {
+        return Promise.reject('Missing SendGrid env')
```
MattiaPrimavera, line 11: pls consider throwing an error here, instead of returning a `reject` promise

functions/src/notifications/sendEmail.ts
```diff
@@ -0,0 +1,41 @@
...
+        to: params.email,
+        from: `InstantVisio <${params.emailFrom}>`,
+        subject: subject,
+        text: messageBody,
+        ip_pool_name: params.sendGridCredentials.ip_pool_name,
+    }
+
+    try {
+        const result = await sgMail.send(msg)
+        const response = result && result[0]
```
MattiaPrimavera, line 25: not sure, but would the following give the same result ?

```js
const [emailSendResult] = sgMail.send(msg)
```

functions/src/notifications/sendNotifications.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export const sendNotifications = async (
+    destinations: InvitationDestination[],
+    notificationContent: NotificationContent,
+    ovhCredentials?: OVHCredentials,
+    sendGridEnv?: SendGridEnv
+): Promise<SendNotificationsResult> => {
+    // Just some quick prefilters
+    const emailDestinations = destinations.filter(
+        (dest) => dest.email && dest.email.length > 4
```
MattiaPrimavera, line 21: may we add this [prefer-optional-chain](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/prefer-optional-chain.md) linter rule ?

functions/src/notifications/sendNotifications.ts
```diff
@@ -0,0 +1,73 @@
...
+                    email: emailDest.email,
+                    ovhCredentials: ovhCredentials,
+                    emailFrom: appEnv.emailFrom,
+                })
+                emailsSent.push(emailDest.email)
+            } catch (error) {
+                console.error(error)
+            }
+        }
+    }
```
MattiaPrimavera, line 49: pls consider refactoring this function in a `sendEmail` function

functions/src/notifications/sendNotifications.ts
```diff
@@ -0,0 +1,73 @@
...
+                    lang: smsDest.lang,
+                    phone: smsDest.phone,
+                    sendGridCredentials: sendGridEnv,
+                })
+                smssSent.push(smsDest.phone)
+            } catch (error) {
+                console.error(error)
+            }
+        }
+    }
```
MattiaPrimavera, line 67: same as above but for `sendSms` function

MattiaPrimavera (changes requested):
Nice work! Just a few feedbacks, there is some code styling we should find an agreement on (or discuss about :))

I did not leave a comment on all the occurrences of the following cases. So if we agree on the code styling preference, pls give a look at the PR as a whole to be aligned with the code styling we choose :)

- prefer destructuring to increase readability and avoid duplication -> I propose to enforce through linter
- prefer optional-chain instead of AND logical expressions -> I propose to enforce through linter
- consider directly returning a value or throwing an error in an async function (to avoid using `Promise.resolve` or `Promise.reject`)

functions/src/api/v1/invite/inviteParticipants.ts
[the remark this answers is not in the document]
HugoGresse: for testing for example, we could have test account for OVH and twilio. I'll move it so the same file though and export two function.

functions/src/api/v1/invite/inviteParticipants.ts
```diff
@@ -0,0 +1,144 @@
...
+ *         description: missing authorization bearer token
+ *       403:
+ *         description: token not valid
+ *       404:
+ *         description: room does not exist
+ *       412:
+ *         description: authorization header present but not formatted correctly
+ */
+export const inviteParticipants = wrap(async (req: Request, res: Response) => {
+    const roomId = req.params.roomId
```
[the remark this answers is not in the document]
HugoGresse: Destructuring has a cost for executing, memory and also reading (in my case a least). 
It's useful when needed (line 88 for eg), not in this case

functions/src/api/v1/invite/inviteParticipants.ts
```diff
@@ -0,0 +1,144 @@
...
+            return {
+                ...dest,
+                lang,
+                country,
+            }
+        }
+    )
+
+    // TODO :
+    // - check pricing maybe ot log email & sms sent on user account
```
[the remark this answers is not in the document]
HugoGresse: I mean we'll store the amount of email sent bu the user to track his usage.

functions/src/firebase/env.ts
HugoGresse: Destructuring has a cost though: https://codeburst.io/es6s-function-destructuring-assignment-is-not-free-lunch-19caacc18137

functions/src/notifications/sendNotifications.ts
HugoGresse: I agree, you can open another PR with the rule :p

functions/src/api/v1/invite/inviteParticipants.ts
```diff
@@ -0,0 +1,144 @@
...
+import { getAppEnv, getOVHEnv, getSendGridEnv } from '../../../firebase/env'
+import { NotificationContent } from '../../../types/Notification'
+import { sendNotifications } from '../../../notifications/sendNotifications'
+import { UID } from '../../../types/uid'
+
+/**
+ * @swagger
+ * /v1/rooms/{roomId}/sendInvites:
+ *   post:
+ *     description: Send an invitation to one or many participants, via email and/or sms. It can combine email and sms destinations, as well as specific lang for each participants. Email & phone can be used at the same time, email will be sent first. <br/> Default lang is "en". <br/> Available languages are en, fr, de, es, gr (el), hu, it, ro. <br/> Country are defaulted to "fr" if not supplied, it improve the phone umber parsing success rate, though it is already good by default (it use libphonenumber-js). <br/> Email and phone numbers are not saved in InstantVisio databases, not logged. Phone number are stored for a maximum of 24 hours on OVH Telecom (CRON run to clear them up every 12h), while emails logs are kept on SendGrid (email) without being able to erase those logs.
```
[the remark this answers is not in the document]
HugoGresse: Country is used for phone verification, there is no default en country so I've defaulted to fr which in our cases will have a better success rate than other as the default value. Regarding lang, en being the most spoken, it seems a fair choice.

functions/src/api/v1/invite/inviteParticipants.ts
```diff
@@ -0,0 +1,144 @@
...
+    const notificationSentResult = await sendNotifications(
+        invitationsDestinations,
+        notificationContent,
+        ovhEnv,
+        sendGridEnv
+    )
+
+    if (
+        notificationSentResult.emailsSent.length === 0 &&
+        notificationSentResult.smssSent.length === 0
```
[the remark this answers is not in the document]
HugoGresse: detructuring ok, having a function that check length of two array is not useful and

functions/src/notifications/sendNotifications.ts
HugoGresse: not possible here I think

functions/src/api/v1/invite/inviteParticipants.ts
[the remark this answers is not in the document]
MattiaPrimavera: cost in performance you mean ? I think performance cost is negligible and readability is increased since the line is shorter

```js
const {roomId} = req.params
```

and there is no `roomId` repetition on both sides of the expression. Of course the benefits are clearer when the destructuring includes more than one property, or when the property is used in multiple spots (avoids code duplication)

functions/src/notifications/sendNotifications.ts
MattiaPrimavera: you should be able to go with:

```js
(dest) => dest?.email?.length > 4
```

HugoGresse: cannot do

HugoGresse:
It's ok to next review guys! Hopping to merge tomorrow.

PR #316, opened 2020-12-11
Use safari view controller and InAppBrowser (iOs)

### LINKED ISSUE #307: Use safari-view-controller plugin for Twilio video calls
`safari-view-controller` plugin will be used to integrate Twilio video calls instead of native integration (for `iOs` platform only) for the moment

MattiaPrimavera:
- add plugins for premium video call on `iOs`
- run `ionic cap sync` to add all needed native platform files
- add `safari-view-controller` service file to use the plugin
- add `platform` service file to recognise which device the app is running on

fix #307


src/services/safari-view-controller.ts
```diff
@@ -0,0 +1,29 @@
...
+        SafariViewController.show({
+            url,
+            hidden: false,
+            animated: false,
+            transition: 'curl',
+            enterReaderModeIfAvailable: true,
+            tintColor: '#ff0000',
+        }).subscribe(
+            (result: any) => {
+                if (result.event === 'opened') console.log('Opened')
```
ntchou, line 20: is it necessary ? May be we should do this `console.log(result.event)`

MattiaPrimavera: we don't actually use that indeed, at least for the moment. It comes from the example code

ntchou (approved):

src/services/platform.ts
```diff
@@ -0,0 +1,3 @@
+import { isPlatform } from '@ionic/react'
+
+export const isIos = () => isPlatform('ios')
```
HugoGresse, line 3: not used?

src/services/safari-view-controller.ts
```diff
@@ -0,0 +1,25 @@
...
+        SafariViewController.show({
+            url,
+            hidden: false,
+            animated: false,
+            transition: 'curl',
+            enterReaderModeIfAvailable: true,
+            tintColor: '#ff0000',
+        }).subscribe(
+            (result: any) => console.log(result.event),
+            (error: any) => console.error(error)
```
HugoGresse, line 20: maybe fallback to openWithNativeBrowser on error?

HugoGresse (approved):

src/services/platform.ts
MattiaPrimavera: yep, only added the tools for the moment and tested locally. Need to find the best spot where to use this

src/services/safari-view-controller.ts
MattiaPrimavera: Since this is an observable, the `subscribe` method could be passed several errors during the lifespan of the `safari-view-controller` window, which means we would open several native browser windows in this case

Issue #174, opened 2020-05-02
Node version (nvm)


MattiaPrimavera:
Is there a particular reason to use two different node versions for firebase functions and the front app ?
- functions => `v8.15.0`
- front => `v10.17.0`

Abibibi:
Node 8 is needed for Firebase functions

MattiaPrimavera:
Yep but from [functions docs](https://firebase.google.com/docs/functions/get-started#set-up-node.js-and-the-firebase-cli):

> You'll need a Node.js environment to write functions, and you'll need the Firebase CLI to deploy functions to the Cloud Functions runtime. Node.js versions 8 and 10 are supported. For installing Node.js and npm, Node Version Manager is recommended.

HugoGresse:
Node 10 on GCP is in "beta" but we could indeed switch to 10 on both without issue.

MattiaPrimavera:
Which part of our service is using `GCP` ? Well if it's beta, better not to change that for production ... but the docs I linked seem saying it's supported for `firebase functions`, not beta :)

HugoGresse:
Firebase run on GCP. so Functions are Google Cloud Functions under the shell. 
You can probably make another pr to change the node version on functions and remove the .nvmrc.

PR #317, opened 2020-12-12
API Invite review

HugoGresse:
- change subscription model
- remove .nvmrc for functions & yarn.lock, closing #174


functions/.nvmrc
```diff
@@ -1 +0,0 @@
-v10.22.1
```
MattiaPrimavera, line 1: I guess you thought of having a single `.nvmrc` file for this project, which is a nice idea. Although, since the front is using node `v10.17.0` (little difference), only removing this file causes a `regression` in the node version used for functions

MattiaPrimavera (commented):
There is a few unanswered conversations [here](https://github.com/Instant-Visio/InstantVisio-WebApp/pull/296)

functions/src/types/UserData.ts
```diff
@@ -4,8 +4,10 @@ import { UID } from '../types/uid'
 
 export interface UserData {
     id: UID
-    subscriptionActive: boolean
-    quotaReached: boolean
+    subscription: {
```
MattiaPrimavera, line 7: no more need of `history` and `type` fields ?

functions/.nvmrc
HugoGresse: yes, we don't care much though :D

functions/src/types/UserData.ts
HugoGresse: it's not used currently, and not populated automatically so yeah.

functions/.nvmrc
HugoGresse: which cause which issue?

Issue #303, opened 2020-12-08
Linter enhancements


MattiaPrimavera:
- `tslint` is deprecated as stated [here](https://medium.com/palantir/tslint-in-2019-1a144c2317a9) -> move `functions` to `eslint`
- we now use `typescript` on both back and front-end, which means we do wanna have a consistent code style
- since some of the rules are specific to the front-end or back-end only, we can create a base config and extend it to make it more specific
- add linter as a git commit hook and as a build prescript

MattiaPrimavera:
We may prefer using `typescript-eslint`, for the [differences with babel-eslint](https://github.com/typescript-eslint/typescript-eslint)

MattiaPrimavera:
Add extra linter rules:
- uniform way of declaring React components

PR #324, opened 2020-12-13
Invite Participants by email basic form

### LINKED ISSUE #315: Send email/sms invite
- add **add participant** button + UI
- enhance `api` network layer to invite participants

MattiaPrimavera:
- basic participant invite (only allows inserting one or several emails separated by space)
- "placeholder" translations
- network layer to invite the participant
- one generic error message in case of errors
- `Add participants` button within premium video call UI
- generic reusable `Modal` window component
- display invite form component within a Modal on click to "Add participant" button

### Missing
- form validation
- add participants also by phone or group
- set the correct language for the invite

fix #315

### DIFF
--- a/src/components/InviteParticipants/InviteForm.tsx
+++ b/src/components/InviteParticipants/InviteForm.tsx
@@ -5,7 +5,7 @@ import styled from 'styled-components'
 import { useTranslation } from 'react-i18next'
 import classNames from 'classnames'
 import Field from '../Form/Field'
-import CallError, { Error } from '../Form/CallError'
+import CallError from '../Form/CallError'
 
 const BootstrapForm = styled(BaseForm)`
     width: 100%;
--- a/src/components/InviteParticipants/InviteForm.tsx
+++ b/src/components/InviteParticipants/InviteForm.tsx
@@ -44,12 +44,8 @@ const FormSubmit = styled.div`
     }
 `
 
-export default function InviteForm({ onSubmit }) {
+export default function InviteForm({ onSubmit, error }) {
     const { t } = useTranslation('add-participants-form')
-    const error: Error = {
-        code: '500',
-        message: t('errors.inviteParticipants'),
-    }
 
     const initialValues = {
         participantAddresses: [],
--- a/src/components/InviteParticipants/InviteForm.tsx
+++ b/src/components/InviteParticipants/InviteForm.tsx
@@ -1,0 +1,118 @@
+import React from 'react'
+import { Formik } from 'formik'
+import { Form as BaseForm, Button } from 'react-bootstrap'
+import styled from 'styled-components'
+import { useTranslation } from 'react-i18next'
+import classNames from 'classnames'
+import Field from '../Form/Field'
+import CallError, { Error } from '../Form/CallError'
+
+const BootstrapForm = styled(BaseForm)`
+    width: 100%;
+`
+
+const FormFields = styled.div`
+    & > * {
+        margin: 0 0 ${({ theme }) => theme.spacing.XXL};
+    }
+
+    &:first-child {
+        margin-top: ${({ theme }) => theme.spacing.XXL};
+    }
+`
+
+const SubmitError = styled.p`
+    color: ${({ theme }) => theme.color.red};
+`
+
+const FormCardHeader = styled.div`
+    & p {
+        &:first-child {
+            text-align: center;
+            font-weight: bold;
+            font-size: ${({ theme }) => theme.font.XL};
+            color: ${({ theme }) => theme.color.orange};
+        }
+    }
+`
+
+const FormSubmit = styled.div`
+    margin-bottom: ${({ theme }) => theme.spacing.XXL};
+    text-align: center;
+    Button {
+        margin: ${({ theme }) => theme.spacing.XXS};
+    }
+`
+
+export default function InviteForm({ onSubmit }) {
+    const { t } = useTranslation('add-participants-form')
+    const error: Error = {
+        code: '500',
+        message: t('errors.inviteParticipants'),
+    }
+
+    const initialValues = {
+        participantAddresses: [],
+    }
+
+    const onValidate = async (values) => {
+        // TODO add validation logic
+        console.log('Values: ', values)
+        return true
+    }
+
+    return (
+        <>
+            <FormCardHeader>
+                <p>{t('title')}</p>
+                <p>{t('description')}</p>
+            </FormCardHeader>
+            <Formik
+                initialValues={initialValues}
+                validateOnBlur={false}
+                validateOnChange={false}
+                validate={onValidate}
+                onSubmit={onSubmit}>
+                {(props) => {
+                    /* eslint-disable react/prop-types */
+                    const { isSubmitting, handleSubmit } = props
+                    return (
+                        <BootstrapForm onSubmit={handleSubmit} noValidate>
+                            <FormFields>
+                                <Field
+                                    name="participantAddresses"
+                                    type="text"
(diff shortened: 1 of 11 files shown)


MattiaPrimavera:
Still WIP, early suggestions appreciated ;)

src/components/InviteParticipants/InviteForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+        code: '500',
+        message: t('errors.inviteParticipants'),
+    }
+
+    const initialValues = {
+        participantAddresses: [],
+    }
+
+    const onValidate = async (values) => {
+        // TODO add validation logic
```
HugoGresse, line 59: using YUP, + create a Field directly linked to Formik to handle error msg display

src/components/InviteParticipants/InviteForm.tsx
```diff
@@ -0,0 +1,118 @@
+import React from 'react'
+import { Formik } from 'formik'
+import { Form as BaseForm, Button } from 'react-bootstrap'
```
HugoGresse, line 3: It's weird to use both Bootstrap and MUI...

src/components/InviteParticipants/InviteParticipants.tsx
```diff
@@ -0,0 +1,50 @@
+import React, { useState } from 'react'
+import InviteForm from './InviteForm'
+import Modal from '../Modal/Modal'
+import Button from '@material-ui/core/Button'
+import { selectToken } from '../../utils/selectors'
+import { useSelector } from 'react-redux'
+import { Api } from '../../services/api'
+
+const InviteParticipants = () => {
```
HugoGresse, line 9: in other files, I've seen exporting without default, maybe add a rules for one way or the others.

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,50 @@
...
+            padding: theme.spacing(2, 4, 3),
+        },
+    })
+)
+
+export default function Modal({ isOpened, children, onClose }) {
+    const classes = useStyles()
+    const [modalStyle] = React.useState(getModalStyle)
+
+    const body = (
```
HugoGresse, line 33: why moving the body in a variable?

src/components/InviteParticipants/InviteParticipants.tsx
```diff
@@ -0,0 +1,50 @@
...
+    const onSubmit = (values, { setSubmitting }) => {
+        const { participantAddresses } = values
+        const addresses = participantAddresses.split(' ').map((email) => ({
+            email,
+            lang: 'fr',
+        }))
+        console.log('On submit called: ', addresses)
+
+        try {
+            const api = new Api(token)
```
HugoGresse, line 22: I believe this should be in an redux action, where we request to store to get the token.

HugoGresse (commented):
As a general guidance, it seems to be a great PR ton introduce common desgin component that will be used in other pages, like a common IVInput, IVButton, etc.

src/components/InviteParticipants/InviteForm.tsx
MattiaPrimavera: - validation logic will be implemented once the participants invite input field will be replaced with the more complex component also handling phone numbers and groups
- I reused `CallError` and only set a generic error message in case the network operation fails

src/components/InviteParticipants/InviteParticipants.tsx
MattiaPrimavera: took note in #303

src/components/InviteParticipants/InviteForm.tsx
MattiaPrimavera: Created #328 for this one

src/components/Modal/Modal.tsx
MattiaPrimavera: that comes from [material-ui SimpleModal](https://material-ui.com/components/modal/#simple-modal)

src/components/InviteParticipants/InviteParticipants.tsx
MattiaPrimavera: and I can set the `api` instance within the `store` ?
It would be awesome indeed to instance this class after obtaining a token and make it available to the rest of the application

src/reducers/rootReducer.ts
```diff
@@ -1,24 +1,44 @@
...
-    action: SetTokenAction
+    action: ActionTypes
 ): AppState => {
     switch (action.type) {
         case SET_TOKEN:
             return {
                 ...state,
                 token: action.payload.token,
             }
+        case SET_ROOM_ID:
```
HugoGresse, line 32: using immer js, we could do: 

```
    ({type, payload}) => {

        switch (action.type) {
            case SET_TOKEN:
                draft.token = action
           default:
                return
        }
}
```

src/reducers/rootReducer.ts
```diff
@@ -1,24 +1,44 @@
...
 const initialState = {
     token: null,
+    roomId: null,
+    hostName: null,
 }
 
 const rootReducer = (
     state: AppState = initialState,
-    action: SetTokenAction
+    action: ActionTypes
```
HugoGresse, line 24: as we increase complexity, I think we need to hury up moving out of a rootReducer with all the logic and start splitting the state into multiple one: `authReducer` (or user), `roomReducer`, etc

HugoGresse (changes requested):

src/reducers/rootReducer.ts
MattiaPrimavera: I'll look into `immerjs`

src/components/InviteParticipants/InviteParticipants.tsx
MattiaPrimavera: - the user token is already stored in `redux` (ok to put it in a parent `user` or `token` object
- I think async actions allow to use `dispatch` in a single async function, but I'm not using dispatch here

src/reducers/rootReducer.ts
MattiaPrimavera: #333

src/reducers/rootReducer.ts
MattiaPrimavera: #334

src/components/InviteParticipants/InviteParticipants.tsx
HugoGresse: I understand dispatch is not used, but the invitation of users should be an action dispatched from the component. Even if it does not modify the local or global state.

MattiaPrimavera: why ? In that case, I have the impression that every async function used in a component might be turned into an `async action`

MattiaPrimavera: now that I think about this, if I use a single loader for the app (that overlays the whole UI), I might dispatch the action to display the loader, and thus from within the `inviteParticipants` method, and thus I see a perfect case to move that into an async action ;)

HugoGresse: Yes exactly
- have the same architecture for all component: they want to trigger an action on something, go through redux
- provide better tracking of what is going through within the app with Redux tooling
- separate Component from business side: Component are responsible for the UI and routing, action manage async behavior, network request and business rules. The component just want to invite users, it doesn't care how to find the token, the action will do it for him.

MattiaPrimavera: #336

HugoGresse (approved):

ntchou (approved):

PR #329, opened 2020-12-13
API: get users details route

HugoGresse:
New route to get user details


functions/src/api/swaggerDefinition.ts
```diff
@@ -15,6 +15,29 @@ export const swaggerDefinition = {
...
+                            isActive: 'boolean',
+                            isQuotaReached: 'boolean',
+                        },
+                    },
+                    sentEmails: {
+                        type: 'integer',
+                    },
+                    sentSMSs: {
+                        type: 'integer',
+                    },
```
MattiaPrimavera, line 35: pls consider regrouping `sentEmails` & `sentSMSs` properties in a parent field, like `stats` or `statistics` ?

functions/src/api/swaggerDefinition.ts
```diff
@@ -15,6 +15,29 @@ export const swaggerDefinition = {
...
+                    updatedAt: {
+                        type: 'integer',
```
MattiaPrimavera, line 37: a timestamp is stored here? Is `integer` the best type ?

functions/src/api/v1/users/getUser.ts
```diff
@@ -0,0 +1,57 @@
...
+ *             }
+ *             schema:
+ *               type: object
+ *               $ref: '#/components/schemas/User'
+ *       401:
+ *         description: missing authorization bearer token
+ *       403:
+ *         description: authorization header present but not valid
+ *       412:
+ *         description: authorization header present but not formatted correctly
```
MattiaPrimavera, line 36: pls consider simplifying the two above with:
- `invalid authorization header`
- `authorization header wrong format`

functions/src/api/v1/users/getUser.ts
```diff
@@ -0,0 +1,57 @@
...
+ */
+export const getUser = async (req: Request, res: Response) => {
+    const currentUid = res.locals.uid
+    const requestedUserId = req.params.userId
+
+    if (requestedUserId !== currentUid) {
+        throw new ForbiddenError('Not authorized to get this resource')
```
MattiaPrimavera, line 43: we should discuss again about error translations. If we put those in the frontend side (as I'd suggest), we'd probably need to insert a `string` error code to uniquely identify a translation

functions/src/api/v1/users/getUser.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    const user = await getUserDb(currentUid)
+
+    res.send({
```
MattiaPrimavera, line 48: wouldn't `res.json()` be more appropriate when sending the response to the client, since it also sets the response type ?

functions/src/api/v1/users/getUser.ts
```diff
@@ -0,0 +1,57 @@
...
+        user: {
```
MattiaPrimavera, line 49: as a general rule, I think it only adds extra complexity to wrap responses with a key, we could directly send back the object here, since it's clear what `getUser` should return

functions/src/types/User.ts
```diff
@@ -0,0 +1,31 @@
...
+import { UID } from './uid'
+
+export interface User {
+    id: UID
+    subscription: {
+        isActive: boolean
+        isQuotaReached: boolean
+    }
+    sentSMSs: number
+    sentEmails: number
```
MattiaPrimavera, line 12: same as above

MattiaPrimavera (changes requested):

functions/src/api/swaggerDefinition.ts
HugoGresse: yes it is, primitive type on swagger.

HugoGresse: but timestamp are not stored in db, they are stored as firestore.Timestamp.

functions/src/api/v1/users/getUser.ts
HugoGresse: ForbiddenError result in a 403, which make it easy to have the error on the frontend only.

functions/src/api/v1/users/getUser.ts
HugoGresse: it seems simplie yes, but what if we want to add some extra field to the response later on? We will have to brake compatibility, make API v2 and so on. here it seems to be a good tradeoff.

functions/src/api/v1/users/getUser.ts
HugoGresse: the response type is send automatically using res.send with an object

HugoGresse:
merging this, need to use it for another PR.

functions/src/api/v1/users/getUser.ts
MattiaPrimavera: Yes the front receives the error code, what I meant is to receive a string ID to find the correct translation on front-end side

functions/src/api/v1/users/getUser.ts
MattiaPrimavera: Are we sure about this? Pls give a look [here](https://www.tutorialspoint.com/difference-between-res-send-and-res-json-in-express-js)

functions/src/api/v1/users/getUser.ts
MattiaPrimavera: I see your point, but I've never seen this practice in my experience. If a `getUser` function, we may add some user related extra fields, which would still be part of the `User` object, so I still think wrapping all API responses  adds complexity. @ntchou ?

MattiaPrimavera (commented):
I saw you needed to merge. I left my review after merge anyway to keep discussing some points

functions/src/api/v1/users/getUser.ts
HugoGresse: open an issue! :D

functions/src/api/v1/users/getUser.ts
HugoGresse: I am, 
Capture d’écran 2020-12-14 à 13 52 16

functions/src/api/v1/users/getUser.ts
MattiaPrimavera: #331

Issue #309, opened 2020-12-09
Redirect to Admin page when the user is admin


MattiaPrimavera:
- only in case the user is not `isAnonymous`
- redirect to admin dashboard after successful login

MattiaPrimavera:
Last point is concerned by #319

PR #330, opened 2020-12-13
[FRONT] #309: redirect to admin dashboard if not anonymous

ntchou:
- ProtectedRoute component
- pending token status in reducer
- wip admin dashboard

login


src/actions/types.ts
```diff
@@ -4,6 +4,7 @@ export const SET_TOKEN = 'SET_TOKEN'
 
 export interface Token {
     token: JWTToken
+    isPending: boolean
```
MattiaPrimavera, line 7: what do you mean with pending ? Could u pls provide some details about this logic ?

src/components/App/ProtectedRoute.tsx
```diff
@@ -0,0 +1,36 @@
...
+    const isPending = useSelector(selectIsPendingToken)
+
+    if (isPending) return null
+
+    return (
+        <Route
+            {...rest}
+            render={(props) => {
+                return !isAnonymous && token ? (
+                    <Component {...props} />
```
MattiaPrimavera, line 27: if you meant to create a generic `ProtectedRoute`, pls mind that an anonymous user is also a logged-in user, and needs to be able accessing `ProtectedRoutes`, the difference is:
- an anonymous user:
    - is redirected to the home page after login
    - cannot create premium video calls (but can join)
- a non-anonymous user:
    - is redirected to the *admin dashboard* after login
    - can create & join premium video calls

MattiaPrimavera: You may rename the component to `PremiumRoute` maybe, and create another one for `ProtectedRoute`

src/components/Login.tsx
```diff
@@ -8,14 +8,14 @@ import { fetchToken } from '../services/fetch-token'
...
-import { selectToken } from '../utils/selectors'
+import { selectIsPendingToken, selectToken } from '../utils/selectors'
 import { useDispatch, useSelector } from 'react-redux'
 
 const uiConfig = {
     // Popup signin flow rather than redirect flow.
     signInFlow: 'popup',
     // Redirect to /signedIn after sign in is successful. Alternatively you can provide a callbacks.signInSuccess function.
-    signInSuccessUrl: '/',
+    signInSuccessUrl: '/admin',
```
MattiaPrimavera, line 18: different case by user type here

src/components/Login.tsx
```diff
@@ -24,8 +24,9 @@ const uiConfig = {
         firebaseuiAuth.AnonymousAuthProvider.PROVIDER_ID,
     ],
     callbacks: {
-        // Avoid redirects after sign-in.
-        signInSuccessWithAuthResult: () => false,
+        signInSuccessWithAuthResult: ({ user }) => {
+            return !user.isAnonymous
```
MattiaPrimavera, line 28: - pls consider leaving the comment since I find the method name not that clear
- is the anonymous user correctly redirected to the `Home` page ?

src/pages/Admin/Dashboard.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+import { IonContent } from '@ionic/react'
+import { useTranslation } from 'react-i18next'
+import styled from 'styled-components'
+
+const InformationBlock = styled.div`
```
MattiaPrimavera, line 6: pls consider removing the `Block` suffix, don't we risk to append this to a lot of components otherwise ?

src/pages/Home/EmulatorLogin.tsx
```diff
@@ -3,10 +3,13 @@ import { Button } from 'react-bootstrap'
 import { Link } from 'react-router-dom'
 import { TEST_ACCOUNTS } from '../../constants'
 import { signInEmulatorEmailPassword } from '../../utils/emulators'
+import { useHistory } from 'react-router-dom'
 
 const [FIRST_TEST_ACCOUNT, SECOND_TEST_ACCOUNT] = TEST_ACCOUNTS
 
 export const EmulatorLogin = ({ authInstance, token }) => {
+    const history = useHistory()
+    const isAnonymous = authInstance.currentUser?.isAnonymous
```
MattiaPrimavera, line 12: we may wanna include this information in redux, dispatching it after successful login

MattiaPrimavera (changes requested):

src/components/App/ProtectedRoute.tsx
ntchou: It was a little bit confusing. Thanks for the explanations. It's ok now

src/pages/Admin/Dashboard.tsx
ntchou: It's a copy/paste from `WelcomeCall`. I'll fix it

src/pages/Home/EmulatorLogin.tsx
ntchou: you're right

src/components/Login.tsx
ntchou: I did it related to this
https://github.com/Instant-Visio/InstantVisio-WebApp/pull/330#discussion_r542139282

src/actions/types.ts
ntchou: The token is not filled yet. We're listening on `onAuthStateChanged` function to fill it. `isPending` lets me know when the token is filled. It usefull to avoid flash effects

src/components/Login.tsx
```diff
@@ -43,7 +45,7 @@ const Login = () => {
 
     const login = useCallback(
         (token: JWTToken) => {
-            dispatch(actions.setToken(token))
+            dispatch(actions.setToken(token, false))
```
HugoGresse, line 48: cannot we import just the action we eed to use rather than importing the whole file?

src/reducers/rootReducer.ts
```diff
@@ -15,9 +17,11 @@ const rootReducer = (
 ): AppState => {
     switch (action.type) {
         case SET_TOKEN:
+            const { payload } = action
             return {
```
HugoGresse, line 21: we could also use immer js for immutability

src/utils/selectors.ts
```diff
@@ -1,6 +1,12 @@
+import { AppState } from '../reducers/rootReducer'
 import { createSelector } from 'reselect'
 
 export const selectToken = createSelector(
-    (state: any) => state.token,
+    (state: AppState) => state.token,
     (token) => token
 )
+
+export const selectIsPendingToken = createSelector(
```
HugoGresse, line 9: do we really need to use reselect for that simple getters?

ntchou: you're right. It's overkill just for simple getter

src/pages/Admin/Dashboard.tsx
MattiaPrimavera: Sorry, I just thought about that, not blocking anyway

src/utils/selectors.ts
MattiaPrimavera: on my side I did since `selectToken` is reused, even if the selector is a simple one. I guess that's also a point to consider

src/actions/types.ts
MattiaPrimavera: wouldn't that be the same to verify `if(!token)` to achieve this ?

src/utils/selectors.ts
HugoGresse: well, the goal of reselect if to optimize for big selectors with map/filters and aggregation. 
Here, in term of performance, I bet destructuring is more impactful than properties access on object :D

MattiaPrimavera: I used that for [memoizing selectors](https://react-redux.js.org/api/hooks#using-memoizing-selectors)

src/actions/types.ts
ntchou: Nop

You're token is not initialized yet. Let me show with a screencast

ntchou:
Draft:
refactoring

MattiaPrimavera:
- `GitHub` has a button to set the PRs to `draft`
- pls consider asking again for a review when you answer to requested changes
- pls consider using [closing keywords](https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords) and link the issue the PR is handling on the right of GitHub UI

src/actions/types.ts
MattiaPrimavera: I think I understood, I'd be pleased to see a screencast if you have the time, otherwise I'll test it locally

src/components/App/userReducer.ts
```diff
@@ -1,30 +1,57 @@
...
 export interface User {
     token: JWTToken
     userId: UID
+    isAnonymous: boolean
 }
+
 export interface UserState {
     user: User
+    isLoading: boolean
+    error: null | unknown
```
MattiaPrimavera, line 16: (non-blocking) I think we'd rather set default values to empty string instead of having `| null` for non-initialised states

MattiaPrimavera (changes requested):
Good job :)
- I need to catch up with `immerjs`
- I appreciated the concepts of `SignIn` and `SignOut` instead of using `hasToken`
- some conversations still need to be solved
- nice making the Login component cleaner

src/components/App/userReducer.ts
ntchou: I set to null because i had a case where the firebase error was an objet

MattiaPrimavera (approved):

HugoGresse (approved):

PR #344, opened 2020-12-15
Create test users

### LINKED ISSUE #319: Create test users
When developing locally, it would be ideal to have a db seed to set up a few test users.

Firebase emulators don't implement a way to `auth:import` & `auth:export` yet, for the `auth` data. This means we have to handle this manually with some `ts` code.

At the moment, I create two accounts from the front-end side (only when using emulator), and I could change related test users data in `onUserCreate` hook.

A better way would be to setup the db through the `Admin SDK`.

MattiaPrimavera:
Fix #319


functions/src/triggers/onUserCreation.ts
```diff
@@ -3,11 +3,49 @@ import * as jsonWebToken from 'jsonwebtoken'
...
+        sentEmails: 0,
+    },
+})
+
+const getUserData = (user: admin.auth.UserRecord) => {
+    let userData
+    if (process.env.FIREBASE_AUTH_EMULATOR_HOST) {
+        switch (user.email) {
+            case TEST_ACCOUNTS.paidUser.email:
+                userData = makeUserData(true, false)
```
HugoGresse, line 29: return directly rather than mutating

functions/src/db/constants.ts
```diff
@@ -7,3 +7,18 @@ export const COLLECTIONS = {
 export const SUBSCRIPTIONS = {
     manual: 'manual',
 }
+
+export const TEST_ACCOUNTS = {
```
HugoGresse, line 11: duplicated file from back and front :(

functions/src/triggers/onUserCreation.ts
MattiaPrimavera: clean code: a function should rather have a single success exit point

HugoGresse: yes but what about immutability?

functions/src/db/constants.ts
MattiaPrimavera: #352

functions/src/triggers/onUserCreation.ts
MattiaPrimavera: I guess it is a compromise to accept in order to increase readability in this case ;)

HugoGresse: what about reducer? they all return directly within a switch

HugoGresse (approved):

functions/src/triggers/onUserCreation.ts
MattiaPrimavera: good point. We may consider [refactoring that](https://medium.com/better-programming/write-cleaner-reducers-1c433a94a166)

PR #362, opened 2020-12-16
API: update rooms routes to support invitation, reminders and room name

HugoGresse:
- refactor inviteParticipants & createReminder to be used without express
- add three new parameters to createRoom to support sending the invitations right away or schedule any number of reminders
- add `name` to roomn on creation and edit
- small swagger refacto. I cannot share all parameters as some routes parameters are required and other not, which is not possible using swagger $ref


HugoGresse:
@MattiaPrimavera @ntchou can I have a review?

functions/src/api/v1/rooms/createRoom.ts
```diff
@@ -101,9 +144,66 @@ export const createRoom = async ({
...
+            const id = await createReminder({
+                roomId,
+                userId,
+                sendAtParameter: sendAt,
+                hostName,
+                destinationsParameter: destinations,
+            })
+            reminderIds.push(id)
+        }
+        return {
```
ntchou, line 188: why don't you return directly `reminderIds` ?

functions/src/api/v1/rooms/createRoom.ts
```diff
@@ -101,9 +144,66 @@ export const createRoom = async ({
...
+    })
+}
+
+type ProcessDestinationsResponse =
+    | InviteParticipantsResponse
+    | {
+          reminderIds: ReminderId[]
+      }
+
+const parseSendsAt = (sendsAt: string): string[] => {
```
ntchou, line 207: Maybe ?
```suggestion
const parseSendsAt = (sendsAt: string | Array<unknow>): string[] => {
```

functions/src/api/v1/rooms/editRoom.ts
```diff
@@ -47,17 +40,20 @@ export const editRoom = wrap(async (req: Request, res: Response) => {
         id: roomId,
     }
 
-    if (req.body.password) {
-        dataToEdit['password'] = req.body.password
+    if (req.body.password && req.body.password.length > 0) {
```
ntchou, line 43:
```suggestion
    if (req.body?.password?.length) {
```

or if you prefer `req.body?.password?.length > 0`

functions/src/api/v1/rooms/editRoom.ts
```diff
@@ -47,17 +40,20 @@ export const editRoom = wrap(async (req: Request, res: Response) => {
...
+        dataToEdit.password = req.body.password
     }
     if (req.body.hideChatbot) {
         dataToEdit['hideChatbot'] = req.body.hideChatbot === 'true'
     }
     if (req.body.startTimestamp) {
-        dataToEdit['startAt'] = Timestamp.fromMillis(
+        dataToEdit.startAt = Timestamp.fromMillis(
             parseInt(req.body.startAt) * 1000
```
ntchou, line 51:
```suggestion
            +req.body.startAt * 1000
```

functions/src/api/v1/invite/inviteParticipants.ts
```diff
@@ -59,16 +60,40 @@ import { increment } from '../../../firebase/firebase'
...
+    async (req: Request, res: Response) => {
+        const roomId = req.params.roomId
+        const userId: UID = res.locals.uid
+        const { hostName, destinations } = req.body
+
+        const invitationsSent = await inviteParticipant({
+            roomId,
+            userId,
+            hostName,
+            destinationsParameter: destinations,
```
MattiaPrimavera, line 73: pls consider `JSON.parsing` the destinations here, so that you can directly forward the `destination` parameter to `inviteParticipant` function

functions/src/api/v1/reminders/createReminder.ts
```diff
@@ -58,15 +61,41 @@ import { assertTimestampInFuture } from './assertTimestampInFuture'
...
+export const createReminder = async ({
+    roomId,
+    userId,
+    sendAtParameter,
+    hostName,
+    destinationsParameter,
+}: {
+    roomId: RoomId
+    userId: UID
+    sendAtParameter: string
```
MattiaPrimavera, line 91: as a general rule, I would avoid having to suffix variables with `Parameter` and refactor

functions/src/api/v1/reminders/createReminder.ts
```diff
@@ -58,15 +61,41 @@ import { assertTimestampInFuture } from './assertTimestampInFuture'
...
+    destinationsParameter: string
+}): Promise<ReminderId> => {
     await assertRightToEditRoom(roomId, userId)
 
-    const body = req.body
-    const sendAt = parseInt(body.sendAt) * 1000
-    const hostName = req.body.hostName
-    const destinations = JSON.parse(body.destinations || '[]')
+    const sendAt = parseInt(sendAtParameter) * 1000
+    const destinations = JSON.parse(destinationsParameter || '[]')
```
MattiaPrimavera, line 98: `JSON.parse` should be wrapped in order to provide the API consumer with a more specific error (like wrong arguments format)

functions/src/api/v1/rooms/createRoom.ts
```diff
@@ -7,40 +7,71 @@ import { assertNewRoomCreationGranted } from '../subscription/assertNewRoomCreat
...
+} from '../invite/inviteParticipants'
 
 /**
  * @swagger
  * /v1/rooms/new:
  *   post:
- *     description: Create a new room. This will generate a random room id (9 random a-Z 0-9 chat) and a random password if none provided. The room will have an infinite lifetime though it will probably change in the future. <br/><br/>To schedule a room, use this route and set the startTimestamp field, it will not prevent the meeting to start before of after and will be used to fill the date on the UI & reminders.
+ *     description: Create a new room. This will generate a random room id (9 random a-Z 0-9 chat) and a random password if none provided. The room will have an infinite lifetime though it will probably change in the future. <br/><br/>To schedule a room, use this route and set the startTimestamp field, it will not prevent the meeting to start before of after and will be used to fill the date on the UI & reminders. <br/><br/>This route also allow sending invitation to join it right away or setup reminders for the future. For this, always supply hostName & destinations, and use sendsAt only for reminders. This route will either send invitation or schedule reminder, not both.
```
MattiaPrimavera, line 21: `This route also allow` -> `allows`

functions/src/api/v1/rooms/createRoom.ts
```diff
@@ -7,40 +7,71 @@ import { assertNewRoomCreationGranted } from '../subscription/assertNewRoomCreat
...
+ *         examples:
+ *            mixed:
+ *                summary: Multiple sendAt dates
+ *                value: '[1708118298, 1808118298]'
  *         type: integer
  *       - $ref: '#/components/parameters/room/hideChatbot'
  *     responses:
  *       201:
- *         description: Room created with success
+ *         description: Room created with success. Depending of the parameters, will either have a list of created reminders ids OR the emails & SMSs sent list.
```
MattiaPrimavera, line 62: `Depending of` -> `Depending on`

MattiaPrimavera: `will either have` -> `it will either be`

functions/src/api/v1/rooms/createRoom.ts
MattiaPrimavera: This is a point we should discuss, since it is a convention used over probably all API responses. In my opinion we should not to wrap the API response with a key, since the return value is already made clear by the `swagger` and the `endpoint` url. This would cause having a double wrapping: on one side we wrap the response on the API, on the other side we'll have to unwrap it on the client side

functions/src/api/v1/rooms/createRoom.ts
MattiaPrimavera: why unknown and not any ?

functions/src/api/v1/rooms/createRoom.ts
```diff
@@ -101,9 +144,66 @@ export const createRoom = async ({
...
+}
+
+type ProcessDestinationsResponse =
+    | InviteParticipantsResponse
+    | {
+          reminderIds: ReminderId[]
+      }
+
+const parseSendsAt = (sendsAt: string): string[] => {
+    return Array.isArray(sendsAt) ? sendsAt : JSON.parse(sendsAt)
```
MattiaPrimavera, line 208: same as above concerning `JSON.parse`

functions/src/api/v1/rooms/editRoom.ts
MattiaPrimavera: also, I think `> 0` is not needed

MattiaPrimavera (changes requested):
I avoid comments on `destructuring` 'till we don't force that in the linter ;)

functions/src/api/v1/rooms/createRoom.ts
HugoGresse: the route name is `createRoom` it can either return reminderIds or email and sms sent. I wanted to assert that in the response.

functions/src/api/v1/invite/inviteParticipants.ts
HugoGresse: the idea was to regroupe the field format at a single place to allow easier update in the future (like supporting full json payload)

functions/src/api/v1/reminders/createReminder.ts
HugoGresse: it conflict with the var inside, how would you name it?

functions/src/api/v1/invite/inviteParticipants.ts
MattiaPrimavera: does this mean that:

```js
const invitationsSent = await inviteParticipant({
  roomId,
  userId,
  hostName,
  destinations
})  
```

would not fit your purpose ?

functions/src/api/v1/reminders/createReminder.ts
MattiaPrimavera: seems your handling `seconds` and `millis`, what about `sendAtSecs` & `sendAtMillis` ?

functions/src/api/v1/reminders/createReminder.ts
```diff
@@ -58,15 +62,41 @@ import { assertTimestampInFuture } from './assertTimestampInFuture'
...
+    hostName: string
+    destinationsParameter: string
+}): Promise<ReminderId> => {
     await assertRightToEditRoom(roomId, userId)
 
-    const body = req.body
-    const sendAt = parseInt(body.sendAt) * 1000
-    const hostName = req.body.hostName
-    const destinations = JSON.parse(body.destinations || '[]')
+    const sendAt = parseInt(sendAtParameter) * 1000
```
MattiaPrimavera, line 98: if we were handling `Content-Type: json` here, wouldn't also all the `parseInt` calls not be needed anymore ?

MattiaPrimavera (changes requested):
Is `JSONParse` file part of this PR ? I can't see it ^^

functions/src/api/v1/reminders/createReminder.ts
HugoGresse: 🤫

PR #368, opened 2020-12-17
API: startingAfter query params for getRooms

HugoGresse:
- wrap getRooms to catch promise error (as other routes)
- add query params
-


functions/src/db/RoomDao.ts
```diff
@@ -28,23 +28,35 @@ export class RoomDao {
...
-        const query = await db
+        let query = await db
             .collection(COLLECTIONS.rooms)
             .where('uid', '==', userId)
-            .orderBy('createdAt', 'desc')
 
-        const results = await query.get()
+        if (startingAfter) {
+            query = query
+                .where('startAt', '>=', new Date(startingAfter * 1000))
```
MattiaPrimavera, line 40: why not only `>` since we call it `startAfter` ?

MattiaPrimavera (approved):

PR #372, opened 2020-12-18
Add PermissionVideoAudio.tsx

NicolasHov:
Screenshot 2020-12-18 at 17 22 29


src/pages/PermissionVideoAudio/PermissionVideoAudio.tsx
```diff
@@ -0,0 +1,55 @@
...
+    text-align: center;
+`
+
+function ArrowTopLeft(props) {
+    return (
+        <SvgIcon {...props}>
+            <path
+                fill="currentColor"
+                color="#C3CFD9"
+                d="M19,17.59L17.59,19L7,8.41V15H5V5H15V7H8.41L19,17.59Z"
```
HugoGresse, line 35: No icon available in any pack?

src/pages/PermissionVideoAudio/PermissionVideoAudio.tsx
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export default function PermissionVideoAudio() {
+    const { t } = useTranslation('permissions-audio-video')
+
+    return (
+        <IonContent style={{ '--background': 'white' }}>
+            <InformationBlock>
+                <CenteredText>
+                    <ArrowTopLeft style={{ fontSize: 300 }} />
```
HugoGresse, line 48: ok, so a little precision on this. Depending of the browser AND the OS, the permission could be ask either on the top left, on at the bottom. You may want to check at list chrome on desktop and ios and android, and manage this case. For the other context, do not display any arrow.

src/pages/PermissionVideoAudio/PermissionVideoAudio.tsx
NicolasHov: Found all those icons in community but I needed a bigger size than that. Is there a simplier way ? It would be nice indeed

HugoGresse: change either the font size, the size attribute on the icon itself or a manual fixed size in css. They generally offer of on those ways.

src/pages/PermissionVideoAudio/PermissionVideoAudio.tsx
NicolasHov: ok ! it makes sense thanks

ntchou (approved):

PR #391, opened 2020-12-22
API: Implement auto twilio room re-creation if already ended

HugoGresse:
Twilio room end after 5 minutes without participants. This is not something we want if something is scheduled at 14:00 and everyone join at 15:00. Another PR will work on the reminder type room, this adress a direct issue with existing "now" room. 
- new twilioRoomId used to identify a room on Twilio which have a different id that the IV one. This is done in order to have multiple twilio room for a single instantvisio room. This allow recreating a room by anyone who join a room where twilio talled us it was "ended". We don't want definitive ended room as we cannot take into account twilio usage for pricing.

### DIFF
--- a/functions/src/api/errors/HttpError.ts
+++ b/functions/src/api/errors/HttpError.ts
@@ -54,15 +54,15 @@ export class ReminderNotFoundError extends NotFoundError {
     }
 }
 
-export class GoneError extends HttpError {
+class GoneError extends HttpError {
     constructor(message?: string) {
         super(410, message || 'Gone')
     }
 }
 
-export class RoomGoneError extends HttpError {
+export class RoomEndedError extends GoneError {
     constructor() {
-        super(410, 'Room ended')
+        super('Room ended')
     }
 }
 
--- a/functions/src/api/errors/HttpError.ts
+++ b/functions/src/api/errors/HttpError.ts
@@ -54,6 +54,18 @@ export class ReminderNotFoundError extends NotFoundError {
     }
 }
 
+export class GoneError extends HttpError {
+    constructor(message?: string) {
+        super(410, message || 'Gone')
+    }
+}
+
+export class RoomGoneError extends HttpError {
+    constructor() {
+        super(410, 'Room ended')
+    }
+}
+
 export class PreconditionFailedError extends HttpError {
     constructor(message?: string) {
         super(412, message || 'PreconditionFailed')
--- a/functions/src/api/v1/rooms/createRoom.ts
+++ b/functions/src/api/v1/rooms/createRoom.ts
@@ -7,12 +7,19 @@ import { assertNewRoomCreationGranted } from '../subscription/assertNewRoomCreat
 import { RoomId } from '../../../types/Room'
 import { Timestamp } from '../../../firebase/firebase'
 import { RoomDao } from '../../../db/RoomDao'
+import { ReminderId } from '../../../types/Reminder'
+import { createReminder } from '../reminders/createReminder'
+import {
+    inviteParticipant,
+    InviteParticipantsResponse,
+} from '../invite/inviteParticipants'
+import { JSONParse } from '../utils/JSONParse'
 
 /**
  * @swagger
  * /v1/rooms/new:
  *   post:
- *     description: Create a new room. This will generate a random room id (9 random a-Z 0-9 chat) and a random password if none provided. The room will have an infinite lifetime though it will probably change in the future. <br/><br/>To schedule a room, use this route and set the startTimestamp field, it will not prevent the meeting to start before of after and will be used to fill the date on the UI & reminders.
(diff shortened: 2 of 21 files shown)


functions/src/api/v1/rooms/createRoom.ts
```diff
@@ -19,7 +19,7 @@ import { JSONParse } from '../utils/JSONParse'
...
+ *     description: Create a new room. This will generate a random room id (9 random a-Z 0-9 chat) and a random password if none provided. The room will have an a 24 hours lifetime after its startAt time (if not supplied, the creation time). <br/><br/>To schedule a room, use this route and set the startTimestamp field, it will not prevent the meeting to start before of after and will be used to fill the date on the UI & reminders. <br/><br/>This route also allows sending invitation to join it right away or setup reminders for the future. For this, always supply hostName & destinations, and use sendsAt only for reminders. This route will either send invitation or schedule reminder, not both.
```
ntchou, line 22:
```suggestion
 *     description: Create a new room. This will generate a random room id (9 random a-Z 0-9 chat) and a random password if none provided. The room will have a 24 hours lifetime after its startAt time (if not supplied, the creation time). <br/><br/>To schedule a room, use this route and set the startTimestamp field, it will not prevent the meeting to start before or after and will be used to fill the date on the UI & reminders. <br/><br/>This route also allows sending invitation to join it right away or setup reminders for the future. For this, always supply hostName & destinations, and use sendsAt only for reminders. This route will either send invitation or schedule reminder, not both.
```

functions/src/api/v1/rooms/joinRoom.ts
```diff
@@ -65,13 +72,27 @@ export const joinRoom = wrap(async (req: Request, res: Response) => {
     if (room.password !== roomPassword && !isCurrentAdmin) {
         throw new ForbiddenError('Wrong password to join the room')
     }
+    if (room?.status === StatusEnded) {
```
ntchou, line 75: It could be an enhancement to use functions to give room status. For example something like this

```
isStatusEnded(room: Room) => room.status === statusEnded
```

functions/src/api/v1/rooms/service/createTwilioRoom.ts
```diff
@@ -1,15 +1,20 @@
...
-        uniqueName: roomId,
+        uniqueName: twilioRoomId,
         mediaRegion: 'de1',
         type: 'group',
         statusCallback: `https://${domain}/api/v1-private/webhook/twilioStatusCallback`,
     })
-    return room.sid
+    return {
+        sid: room.sid,
+        twilioRoomId: twilioRoomId,
```
ntchou, line 18:
```suggestion
        twilioRoomId,
```

functions/src/api/v1/rooms/service/twilioUtils.ts
```diff
@@ -0,0 +1,5 @@
+import { RoomId } from '../../../../types/Room'
+
+export const formatTwilioRoomId = (roomId: RoomId) => `${roomId}#${Date.now()}`
```
ntchou, line 3: missing return type

functions/src/api/v1/rooms/createRoom.ts
```diff
@@ -120,7 +120,7 @@ export const createRoom = async ({
     const roomPassword =
         roomRequestedPassword || `${~~(Math.random() * 999999)}`
     const roomStartAt = Timestamp.fromMillis(
-        startAt ? parseInt(startAt) * 1000 : ~~(Date.now() / 1000)
+        startAt ? parseInt(startAt) * 1000 : Date.now()
```
ntchou, line 123:
```suggestion
        startAt ? +startAt * 1000 : Date.now()
```

functions/src/api/v1/rooms/joinRoom.ts
HugoGresse: idealy, having room.isStatusEnded will be great too, thus having a real Room class

functions/src/api/errors/HttpError.ts
```diff
@@ -54,6 +54,18 @@ export class ReminderNotFoundError extends NotFoundError {
     }
 }
 
+export class GoneError extends HttpError {
```
MattiaPrimavera, line 57: Is this needed ? I find it not so clear

functions/src/api/errors/HttpError.ts
```diff
@@ -54,6 +54,18 @@ export class ReminderNotFoundError extends NotFoundError {
...
+    constructor(message?: string) {
+        super(410, message || 'Gone')
+    }
+}
+
+export class RoomGoneError extends HttpError {
```
MattiaPrimavera, line 63: I think `RoomEndedError` would be more explicit

MattiaPrimavera (changes requested):
Good job!

functions/src/api/errors/HttpError.ts
HugoGresse: Could be useful for other error but I'll remove the export

Issue #393, opened 2020-12-23
Replace fetch with axios


MattiaPrimavera:
With `fetch` we need to check the response status in order to handle the error. `Axios` is instead `thenable`, thus removing the need to manually check the response status

PR #394, opened 2020-12-23
Set correct room password & join room fixes

### LINKED ISSUE #345: Set correct room password
- use the password when in the join url
- do not use any password in case the premium user wants to create a new room

MattiaPrimavera:
- Set correct room password (previously hard-coded) ->
workaround to handle the admin user navigating to `PremiumVideo` page: from the Dashboard we send him to `/premium-video/room/${roomId}?passcode=admin` since the passcode is required by the original code. When the passcode is `admin` although, it is not gonna be sent to `joinRoom` endpoint, so that a new password will be created on API side
- Hide sign out button for anonymous users
- retry `getJwtToken` (explained in code comment)
- move temporary login UI into a Modal window. Show the modal on login button click
- Add `LoginState` component that subscribes to auth state changes (otherwise the subscription won't occur, since `Login` component is now within a Modal window)
- pave the way to remove `EmulatorLogin` -> emulator login helper buttons moved into `LoginModal`
- handle join room error: "Token must be a string" was displayed
- send `participantName` to `joinRoom` endpoint

Fix #345

### DIFF
--- a/src/actions/getJwtToken.ts
+++ b/src/actions/getJwtToken.ts
@@ -7,19 +7,23 @@ const RETRY = {
     interval: 300,
 }
 
+async function sleep(intervalMillis: number) {
+    return new Promise((resolve) => {
+        setTimeout(function () {
+            resolve(true)
+        }, intervalMillis)
+    })
+}
+
 async function tryAtMost(promise, maxRetries, retryInterval) {
     let retriesNumber = 0
     while (retriesNumber <= maxRetries) {
         try {
+            retriesNumber++
             const result = await promise
             return result
         } catch (err) {
-            retriesNumber++
-            await new Promise((resolve) => {
-                setTimeout(function () {
-                    resolve(true)
-                }, retryInterval)
-            })
+            await sleep(RETRY.interval + RETRY.number ** 1.3)
         }
     }
 
@@ -33,7 +37,7 @@ export const getJwtToken = async (uid: UID): Promise<JWTToken> => {
         })
 
         /*
-            tryAtMost is needed because before a token can exist, firebase functions hook
+            tryAtMost is needed because before a token can exist, Firebase user create trigger
             need to successfully complete execution. Since this call is made right after login,
             the hook might have not finished executing yet
         */
--- a/src/actions/getJwtToken.ts
+++ b/src/actions/getJwtToken.ts
@@ -2,11 +2,46 @@ import { functions } from '../firebase/firebase'
 import { UID } from '../../types/uid'
 import { JWTToken } from '../../types/JWT'
 
+const RETRY = {
+    number: 4,
+    interval: 300,
+}
+
+async function tryAtMost(promise, maxRetries, retryInterval) {
+    let retriesNumber = 0
+    while (retriesNumber <= maxRetries) {
+        try {
+            const result = await promise
+            return result
+        } catch (err) {
+            retriesNumber++
+            await new Promise((resolve) => {
+                setTimeout(function () {
+                    resolve(true)
+                }, retryInterval)
+            })
+        }
+    }
+
+    throw new Error()
+}
+
 export const getJwtToken = async (uid: UID): Promise<JWTToken> => {
     try {
-        const result = await functions.getToken({
+        const getTokenPromise = functions.getToken({
             uid,
         })
+
+        /*
+            tryAtMost is needed because before a token can exist, firebase functions hook
+            need to successfully complete execution. Since this call is made right after login,
+            the hook might have not finished executing yet
+        */
+        const result = await tryAtMost(
+            getTokenPromise,
+            RETRY.number,
+            RETRY.interval
+        )
 
         const jwtToken: JWTToken = result?.data.token
         if (!jwtToken) {
--- a/src/components/App/AuthStateChangedListener/AuthStateChangedListener.tsx
+++ b/src/components/App/AuthStateChangedListener/AuthStateChangedListener.tsx
@@ -4,12 +4,11 @@ import { didSignin } from '../../../actions/userActions'
 import { useDispatch } from 'react-redux'
 import { hideLoginModal } from '../../LoginModal/loginModalActions'
 
-const LoginState = () => {
+const AuthStateChangedListener = () => {
     const dispatch = useDispatch()
 
     useEffect(() => {
         return authInstance.onAuthStateChanged((user) => {
-            console.log('Auth state changed: ', user)
             dispatch(didSignin(user))
             dispatch(hideLoginModal())
         })
@@ -18,4 +17,4 @@ const LoginState = () => {
     return <></>
 }
 
-export default LoginState
+export default AuthStateChangedListener
--- a/src/components/App/LoginState/LoginState.tsx
+++ b/src/components/App/LoginState/LoginState.tsx
@@ -1,0 +1,21 @@
+import React, { useEffect } from 'react'
+import { authInstance } from '../../../firebase/firebase'
+import { didSignin } from '../../../actions/userActions'
+import { useDispatch } from 'react-redux'
+import { hideLoginModal } from '../../LoginModal/loginModalActions'
+
+const LoginState = () => {
+    const dispatch = useDispatch()
+
+    useEffect(() => {
+        return authInstance.onAuthStateChanged((user) => {
+            console.log('Auth state changed: ', user)
+            dispatch(didSignin(user))
+            dispatch(hideLoginModal())
+        })
+    }, [dispatch])
+
+    return <></>
+}
+
+export default LoginState
--- a/src/components/LoginModal/LoginModal.tsx
+++ b/src/components/LoginModal/LoginModal.tsx
@@ -31,36 +31,30 @@ export default function LoginModal() {
     }
 
     return (
-        <>
-            <Modal
-                isOpened={isDisplayed}
-                onClose={() => {
-                    dispatch(hideLoginModal())
-                }}>
-                <Login />
-                {isAuthEmulatorEnabled() && (
-                    <>
-                        <Button
-                            onClick={() =>
-                                signInWithEmailAndPassword(paidUser)
-                            }>
-                            Paid
-                        </Button>
-                        <Button
-                            onClick={() =>
-                                signInWithEmailAndPassword(unpaidUser)
-                            }>
-                            Unpaid
-                        </Button>
-                        <Button
-                            onClick={() =>
-                                signInWithEmailAndPassword(overQuotaUser)
-                            }>
-                            OverQuota
-                        </Button>
-                    </>
-                )}
-            </Modal>
-        </>
+        <Modal
+            isOpened={isDisplayed}
+            onClose={() => {
+                dispatch(hideLoginModal())
+            }}>
+            <Login />
+            {isAuthEmulatorEnabled() && (
+                <>
+                    <Button
+                        onClick={() => signInWithEmailAndPassword(paidUser)}>
+                        Paid
+                    </Button>
+                    <Button
+                        onClick={() => signInWithEmailAndPassword(unpaidUser)}>
+                        Unpaid
+                    </Button>
+                    <Button
+                        onClick={() =>
+                            signInWithEmailAndPassword(overQuotaUser)
+                        }>
+                        OverQuota
+                    </Button>
+                </>
+            )}
+        </Modal>
     )
 }
--- a/src/components/LoginModal/LoginModal.tsx
+++ b/src/components/LoginModal/LoginModal.tsx
@@ -1,0 +1,66 @@
+import React from 'react'
+import { useSelector, useDispatch } from 'react-redux'
+import { selectLoginModal } from './loginModalSelector'
+import { hideLoginModal } from './loginModalActions'
+import Login from '../Login/Login'
+import Modal from '../Modal/Modal'
+import Button from '@material-ui/core/Button'
+import {
+    isAuthEmulatorEnabled,
+    signInEmulatorEmailPassword,
+} from '../../utils/emulators'
+import { TEST_ACCOUNTS } from '../../constants'
+import { authInstance } from '../../firebase/firebase'
+const { paidUser, unpaidUser, overQuotaUser } = TEST_ACCOUNTS
+
+export default function LoginModal() {
+    const { isDisplayed } = useSelector(selectLoginModal)
+    const dispatch = useDispatch()
+    const signInWithEmailAndPassword = async ({ email, password }) => {
+        try {
+            const user = await signInEmulatorEmailPassword(
+                authInstance,
+                email,
+                password
+            )
+            console.log('Signed In with user: ', user)
+        } catch (error) {
+            const { code, message } = error
+            console.log(`Error email auth: ${code}, message: ${message}`)
+        }
+    }
+
+    return (
+        <>
+            <Modal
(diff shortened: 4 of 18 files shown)


src/actions/getJwtToken.ts
```diff
@@ -2,12 +2,47 @@ import { functions } from '../firebase/firebase'
...
+    while (retriesNumber <= maxRetries) {
+        try {
+            const result = await promise
+            return result
+        } catch (err) {
+            retriesNumber++
+            await new Promise((resolve) => {
+                setTimeout(function () {
+                    resolve(true)
+                }, retryInterval)
```
HugoGresse, line 21: We need to increase the delay, maybe more something like this: 
```
setTimeout(, retriesNumber^1.3)
```

src/actions/getJwtToken.ts
```diff
@@ -2,12 +2,47 @@ import { functions } from '../firebase/firebase'
...
+async function tryAtMost(promise, maxRetries, retryInterval) {
+    let retriesNumber = 0
+    while (retriesNumber <= maxRetries) {
+        try {
+            const result = await promise
+            return result
+        } catch (err) {
+            retriesNumber++
+            await new Promise((resolve) => {
+                setTimeout(function () {
```
HugoGresse, line 19: can move to 
```

function sleep(ms) {
  return new Promise(resolve => setTimeout(resolve, ms));
}
```

src/actions/getJwtToken.ts
```diff
@@ -2,12 +2,47 @@ import { functions } from '../firebase/firebase'
...
+
 export const getJwtToken = async (uid: UID): Promise<JWTToken> => {
     try {
-        const result = await functions.getToken({
+        const getTokenPromise = functions.getToken({
             uid,
         })
 
+        /*
+            tryAtMost is needed because before a token can exist, firebase functions hook
```
HugoGresse, line 36: `firebase functions hook` > `Firebase user create trigger need to successfully...` is more precise

src/components/App/LoginState/LoginState.tsx
```diff
@@ -0,0 +1,21 @@
+import React, { useEffect } from 'react'
+import { authInstance } from '../../../firebase/firebase'
+import { didSignin } from '../../../actions/userActions'
+import { useDispatch } from 'react-redux'
+import { hideLoginModal } from '../../LoginModal/loginModalActions'
+
+const LoginState = () => {
```
HugoGresse, line 7:
```suggestion
const AuthStateChangedListener = () => {
```

src/components/LoginModal/loginModalReducer.ts
```diff
@@ -0,0 +1,25 @@
...
+        switch (type) {
+            case HIDE_LOGIN_MODAL:
+                return {
+                    ...draft,
+                    isDisplayed: false,
+                }
+            case SHOW_LOGIN_MODAL:
+                return {
+                    ...draft,
+                    isDisplayed: true,
```
HugoGresse, line 20: if you only mutate one field, you can do draft.isDisplayed directly. 
This where needed for destructuring the whole object in draft, which is not supported by immer.

src/pages/Admin/Dashboard.tsx
```diff
@@ -16,10 +18,19 @@ const InformationBlock = styled.div`
 
 export default function Dashboard() {
     const { t } = useTranslation('admin')
+    const history = useHistory()
+    const startPremiumVideoCall = () => {
+        history.push('/premium-video/room/testRoom?passcode=admin')
```
HugoGresse, line 23: this will not last long, only the first admin will be able to use it, testRoom (roomId) will be already used after.

src/pages/PremiumVideoCall/components/PreJoinScreens/DeviceSelectionScreen/DeviceSelectionScreen.tsx
```diff
@@ -93,12 +92,20 @@ export default function DeviceSelectionScreen({
...
     const handleJoin = async () => {
         setIsLoading(true)
-        const api = new Api(token)
-        const { jwtAccessToken } = await api.joinRoom(roomId, 'test-password') // TODO pass the password variable
-        dispatch(actions.setRoomId(roomId))
-        dispatch(actions.setHostName(name))
-        connect(jwtAccessToken, roomId)
-        setIsLoading(false)
+        try {
+            const twilioVideoToken = await getTwilioVideoToken(
```
HugoGresse, line 96: this could probably be within an action, and dispatch the roomid and hostName within the action.

src/services/api.ts
```diff
@@ -19,9 +19,20 @@ export class Api {
...
-        return this.post(`/rooms/${roomId}/join`, { password })
+        try {
+            const joinResponse = await this.post(`/rooms/${roomId}/join`, {
+                participantName,
+                password,
+            })
+            return joinResponse
+        } catch (err) {
+            throw new Error(
+                'Connection error, maybe the room has not been created yet. Pls try again later'
```
HugoGresse, line 33: use API error message there.

src/services/api.ts
```diff
@@ -44,6 +55,10 @@ export class Api {
             },
             body: data ? JSON.stringify(data) : null,
         })
+
+        if (response.status !== 200 && response.status !== 204) {
```
HugoGresse, line 59: I believe fetch will throw if received error. We can also used `response.ok`,  all http error code >= 200, < 400 are considered valid (FYI)

src/actions/getJwtToken.ts
MattiaPrimavera: why ? Isn't retry with increased delay a retry practice not to make lots of calls, in case a service may be down or can't answer ? In our case we only need the hook to have been executed

src/pages/Admin/Dashboard.tsx
MattiaPrimavera: exactly, this will be fixed with the real room name/id. For the moment it only allows to test a "UI flow" instead of manually changing or pasting the url

src/services/api.ts
MattiaPrimavera: It was *Payment required*, should I use that, so that we'll change the text on the API side if needed ?

src/services/api.ts
MattiaPrimavera: I could print the error result, so I guess it does not throw an error, I also created #393 for this reason. If you agree with #393, this code will be replaced anyway

src/pages/PremiumVideoCall/components/PreJoinScreens/DeviceSelectionScreen/DeviceSelectionScreen.tsx
MattiaPrimavera: How would you call the function to make that clear that it
- gets a video token
- dispatches the hostName
- dispatches the roomName

src/actions/getJwtToken.ts
HugoGresse: yes, the trigger may be cold (because it's a serverless) and the first time can take up to 2 or 3s (hot is aready used within last 10min =  much shorter delay). It's a common practice within mobile apps to increase the retry delay. + I don't know the exact delay of the login trigger.

src/services/api.ts
HugoGresse: I believe we can do that for the time being and check later on for the transation.

src/pages/PremiumVideoCall/components/PreJoinScreens/DeviceSelectionScreen/DeviceSelectionScreen.tsx
HugoGresse: initVideoToken
loadTwilioRoomInfo

we don't need to know what will dispatch the action, just what it's doing overall.

src/services/api.ts
MattiaPrimavera: I replaced as asked

src/pages/PremiumVideoCall/components/PreJoinScreens/DeviceSelectionScreen/DeviceSelectionScreen.tsx
MattiaPrimavera: I think a nice approach could be to create a `connectToRoom` async action, that is dispatched once clicking on `Join`. This action would:
- dispatch the `hostName` & `roomId`
- take care of showing/hiding the loader

We'd thus need to move the loader indicator to redux. I'd treat this in a separate low priority PR although (cleaner code)

src/actions/getJwtToken.ts
```diff
@@ -2,12 +2,51 @@ import { functions } from '../firebase/firebase'
...
+
+async function tryAtMost(promise, maxRetries, retryInterval) {
+    let retriesNumber = 0
+    while (retriesNumber <= maxRetries) {
+        try {
+            retriesNumber++
+            const result = await promise
+            return result
+        } catch (err) {
+            await sleep(RETRY.interval + RETRY.number * 1.3)
```
HugoGresse, line 26: in the last comment i used `^` rahter than multiply, which grealty increase the duration on retry

src/components/App/AuthStateChangedListener/AuthStateChangedListener.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { authInstance } from '../../../firebase/firebase'
+import { didSignin } from '../../../actions/userActions'
+import { useDispatch } from 'react-redux'
+import { hideLoginModal } from '../../LoginModal/loginModalActions'
+
+const AuthStateChangedListener = () => {
+    const dispatch = useDispatch()
+
+    useEffect(() => {
+        return authInstance.onAuthStateChanged((user) => {
```
HugoGresse, line 11: nice

src/pages/PremiumVideoCall/components/PreJoinScreens/DeviceSelectionScreen/DeviceSelectionScreen.tsx
HugoGresse: ok

HugoGresse (changes requested):

src/components/LoginModal/LoginModal.tsx
```diff
@@ -0,0 +1,66 @@
...
+            )
+            console.log('Signed In with user: ', user)
+        } catch (error) {
+            const { code, message } = error
+            console.log(`Error email auth: ${code}, message: ${message}`)
+        }
+    }
+
+    return (
+        <>
```
ntchou, line 34: Fragment is not necessary

src/components/LoginModal/LoginModal.tsx
```diff
@@ -0,0 +1,66 @@
...
+                        <Button
+                            onClick={() =>
+                                signInWithEmailAndPassword(overQuotaUser)
+                            }>
+                            OverQuota
+                        </Button>
+                    </>
+                )}
+            </Modal>
+        </>
```
ntchou, line 64: fragment is not necessary

src/pages/PremiumVideoCall/state/usePasscodeAuth/usePasscodeAuth.ts
```diff
@@ -87,7 +109,9 @@ export default function usePasscodeAuth() {
                         history.replace(window.location.pathname)
                     }
                 })
-                .then(() => setIsAuthReady(true))
```
ntchou, line 90: async/await ?

src/components/App/AuthStateChangedListener/AuthStateChangedListener.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { useEffect } from 'react'
+import { authInstance } from '../../../firebase/firebase'
+import { didSignin } from '../../../actions/userActions'
+import { useDispatch } from 'react-redux'
+import { hideLoginModal } from '../../LoginModal/loginModalActions'
+
+const AuthStateChangedListener = () => {
```
ntchou, line 7: maybe create a hook `useAuthStateChangedListener` instead a react component ?

src/pages/Admin/Dashboard.tsx
```diff
@@ -16,10 +18,19 @@ const InformationBlock = styled.div`
...
+    }
 
     return (
         <IonContent style={{ '--background': 'white' }}>
-            <InformationBlock>{t('dashboard.welcome')}</InformationBlock>
+            <InformationBlock>
+                {t('dashboard.welcome')}
+                <Button onClick={() => startPremiumVideoCall()}>
```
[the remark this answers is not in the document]
MattiaPrimavera: nice suggested change, I'll use in the future. Concerning this specific change, this code will be replaced once integrating the real `DashBoard` page

src/pages/Admin/Dashboard.tsx
```diff
@@ -16,10 +18,19 @@ const InformationBlock = styled.div`
...
+                    Start Call
```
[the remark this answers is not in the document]
MattiaPrimavera: This code will be replaced

src/components/LoginModal/LoginModal.tsx
```diff
@@ -0,0 +1,66 @@
...
+                    dispatch(hideLoginModal())
+                }}>
+                <Login />
+                {isAuthEmulatorEnabled() && (
+                    <>
+                        <Button
+                            onClick={() =>
+                                signInWithEmailAndPassword(paidUser)
+                            }>
+                            Paid
```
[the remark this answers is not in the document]
MattiaPrimavera: These buttons are used to simulate login when using firebase emulator suite locally. No need for translations

src/components/LoginModal/LoginModal.tsx
MattiaPrimavera: What can I return, considering `LoginModal` is a component ?

src/pages/PremiumVideoCall/state/usePasscodeAuth/usePasscodeAuth.ts
MattiaPrimavera: this code is from the original [Twilio Video App project](https://github.com/twilio/twilio-video-app-react). I think we'll have to state in the future if to update all that code in order to meet our code styling guidelines

src/components/App/AuthStateChangedListener/AuthStateChangedListener.tsx
MattiaPrimavera: #409

src/components/LoginModal/LoginModal.tsx
ntchou: Return directly

 ````
return (<Modal>


</Modal>)

````

src/components/LoginModal/LoginModal.tsx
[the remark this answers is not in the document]
ntchou: Ok

src/pages/PremiumVideoCall/state/usePasscodeAuth/usePasscodeAuth.ts
ntchou: Ok

src/pages/Admin/Dashboard.tsx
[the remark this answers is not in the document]
ntchou: Ok

ntchou (approved):

PR #433, opened 2020-12-28
API: createGroup & join group change

HugoGresse:
- update createGroup to provide a groupid and a password
- add joinGroup with user name and password to join an existing group

### DIFF
--- a/functions/src/api/v1/group/createGroup.ts
+++ b/functions/src/api/v1/group/createGroup.ts
@@ -10,6 +10,7 @@ import {
     GroupConflictError,
     GroupNotFoundError,
 } from '../../errors/HttpError'
+import { Member } from '../../../types/Member'
 
 /**
  * @swagger
@@ -73,23 +74,12 @@ export const createGroup = wrap(async (req: Request, res: Response) => {
 
     let groupId: GroupId
     if (requestedGroupId) {
-        try {
-            await GroupDao.get(requestedGroupId)
-            throw new GroupConflictError()
-        } catch (error) {
-            if (error instanceof GroupNotFoundError) {
-                await GroupDao.set(
-                    requestedGroupId,
-                    userId,
-                    name,
-                    password,
-                    members
-                )
-                groupId = requestedGroupId
-            } else {
-                throw error
-            }
-        }
+        groupId = await tryCreateGroupWithPredefinedId(
+            requestedGroupId,
+            userId,
+            password,
+            members
+        )
     } else {
         groupId = await GroupDao.add(userId, name, password, members)
         await GroupDao.update({
@@ -101,3 +91,22 @@ export const createGroup = wrap(async (req: Request, res: Response) => {
         groupId,
     })
 })
+
+const tryCreateGroupWithPredefinedId = async (
+    groupId: GroupId,
+    userId: UID,
+    password: string,
+    members: Member[]
+): Promise<GroupId> => {
+    try {
+        await GroupDao.get(groupId)
+        throw new GroupConflictError()
+    } catch (error) {
+        if (error instanceof GroupNotFoundError) {
+            await GroupDao.set(groupId, userId, name, password, members)
+            return groupId
+        } else {
+            throw error
+        }
+    }
+}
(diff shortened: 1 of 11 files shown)


functions/src/api/v1/group/createGroup.ts
```diff
@@ -47,12 +63,39 @@ export const createGroup = wrap(async (req: Request, res: Response) => {
...
+    }
+
+    let groupId: GroupId
+    if (requestedGroupId) {
+        try {
+            await GroupDao.get(requestedGroupId)
+            throw new GroupConflictError()
+        } catch (error) {
+            if (error instanceof GroupNotFoundError) {
+                await GroupDao.set(
```
MattiaPrimavera, line 81: why `set` ? Shouldn't `create` only do an `add` ?

functions/src/api/v1/group/createGroup.ts
```diff
@@ -47,12 +63,39 @@ export const createGroup = wrap(async (req: Request, res: Response) => {
...
+                    members
+                )
+                groupId = requestedGroupId
+            } else {
+                throw error
+            }
+        }
+    } else {
+        groupId = await GroupDao.add(userId, name, password, members)
+        await GroupDao.update({
```
MattiaPrimavera, line 95: why doesn't `groupDao.add` set the `groupId` as well ?

functions/src/api/v1/group/joinGroup.ts
```diff
@@ -0,0 +1,64 @@
...
+ *       400:
+ *         $ref: '#/components/responses/400'
+ *       401:
+ *         $ref: '#/components/responses/401'
+ *       403:
+ *         $ref: '#/components/responses/403'
+ *       412:
+ *         $ref: '#/components/responses/412'
+ */
+export const joinGroup = wrap(async (req: Request, res: Response) => {
```
MattiaPrimavera, line 42: didn't we already have `/subscribe` endpoint for this ?

MattiaPrimavera (changes requested):

functions/src/api/v1/group/createGroup.ts
HugoGresse: here we want to add a group with a fixed id in advance, requiring set, not add.

functions/src/api/v1/group/createGroup.ts
HugoGresse: the "add" add a new object within a collection, letting firestore pick the right id so we dont have any conflicted id.

functions/src/api/v1/group/joinGroup.ts
HugoGresse: /subscribe is only for the registerToken, but I'll add it optionnaly there too.

functions/src/api/v1/group/createGroup.ts
MattiaPrimavera: (non blocking) but why calling `GroupDao.update` ? Isn't firebase id part of the entity, after the call to `add` ?

functions/src/api/v1/group/createGroup.ts
MattiaPrimavera: (non blocking) I think `createGroup` logic may be expressed in a clearer way

MattiaPrimavera (approved):

functions/src/api/v1/group/createGroup.ts
HugoGresse: the document id is not within the document itself on add, and is retrieve differently where quering the whole collection. It make things simpler to we know the id is always within the document.

functions/src/api/v1/group/createGroup.ts
```diff
@@ -47,14 +64,49 @@ export const createGroup = wrap(async (req: Request, res: Response) => {
...
 })
+
+const tryCreateGroupWithPredefinedId = async (
+    groupId: GroupId,
+    userId: UID,
+    password: string,
+    members: Member[]
+): Promise<GroupId> => {
+    try {
+        await GroupDao.get(groupId)
```
ntchou, line 102: missing `return groupId`

HugoGresse: its below, we throw directly if the await is successful.

PR #437, opened 2020-12-28
[FRONT] #427 hide sign in as guest button

ntchou:


HugoGresse (approved):

MattiaPrimavera:
Did you test if `authInstance.onAuthStateChanged` is called when loggin in anonymously this way ? The app logic relies on that

ntchou:
> Did you test if `authInstance.onAuthStateChanged` is called when loggin in anonymously this way ? The app logic relies on that

Yes. I have a `token` and a `userId` setted

image

PR #442, opened 2020-12-28
Group join enhancement and fixes

MattiaPrimavera:
- `sendRegistrationToken` async action: send the `registrationToken` only after we got a token from the API
- disable `joinGroup` form submit until the form `isValid`
- add some missing translations
- fix: use correct translation for registration error
- use join group endpoint


src/actions/userActions.ts
```diff
@@ -52,6 +54,23 @@ export const setRegistrationToken = (
     },
 })
 
+export const sendRegistrationToken = (
+    t: any,
+    registrationToken: string
+) => async (dispatch, getState) => {
+    const { user: userState } = getState()
+    const { token, userId } = userState.user
```
HugoGresse, line 62: this should be replace by a selector: 
const token = getTokenSelector(getState)

src/actions/userActions.ts
```diff
@@ -52,6 +54,23 @@ export const setRegistrationToken = (
...
+
+    try {
+        const api = new Api(token)
+        await api.addRegistrationToken(userId, registrationToken)
+        dispatch(setRegistrationToken(registrationToken))
+    } catch (e) {
+        console.log('addRegistrationToken error: ', e)
+        dispatch(showErrorMessage(t('errors.register')))
```
HugoGresse, line 70: we could have a generic error in french + a detail in english, so we doesn't need to manage it entirely.

src/components/JoinGroup/JoinGroupForm.tsx
```diff
@@ -13,36 +13,41 @@ import { useSelector } from 'react-redux'
...
+    groupId: string
+    groupPassword: string
+    displayName: string
 }
 
 const JoinGroupForm = ({ onFormSubmit }) => {
     const { t } = useTranslation('join-group-form')
     const token = useSelector(selectToken)
 
+    const validateForm = (values) => {
```
HugoGresse, line 25: still not yup, thus needing manual validatio which is not helping reducing dev time

HugoGresse (changes requested):

src/actions/userActions.ts
MattiaPrimavera: even if I needed the state to also access the `userId` variable ?

src/actions/userActions.ts
MattiaPrimavera: For the moment I've put both translations I think, but I'm not against your idea

src/components/JoinGroup/JoinGroupForm.tsx
MattiaPrimavera: I copy/pasted from `AdminDashboard`, I didn't study yup yet

src/actions/userActions.ts
HugoGresse: another selector with it so

src/components/JoinGroup/JoinGroupForm.tsx
```diff
@@ -13,36 +13,41 @@ import { useSelector } from 'react-redux'
...
+        const errors: Partial<JoinGroupValues> = {}
+        const requiredFields = ['groupId', 'groupPassword', 'displayName']
+        for (const requiredField of requiredFields) {
+            if (!values[requiredField]) {
+                errors[requiredField] = 'Required'
```
HugoGresse, line 30: maybe in french?

HugoGresse (approved):

PR #451, opened 2020-12-29
API: seed route creating user

HugoGresse:
GET `http://localhost:5050/instantvisio-dev/us-central1/api/api/v1-tests/seeds/`

This help local testing using firebase emulators: functions and auth
- create three users (paid, free, overquota)
- onUserCreation from Mattia creation already set the firestore data directly
- return the info
- add "toto" registrationTokens to showcase firestore usage (for future improvement)

### DIFF
--- a/functions/src/api/utils/seeds/seeds.ts
+++ b/functions/src/api/utils/seeds/seeds.ts
@@ -1,0 +1,39 @@
+import { Request, Response } from 'express'
+import { wrap } from 'async-middleware'
+import { seedAuth } from './authSeeds'
+import { seedFirestore } from './seedFirestore'
+import { TEST_ACCOUNTS } from '../../../db/constants'
+
+export const seeds = wrap(async (req: Request, res: Response) => {
+    const authEmulatorUrl = 'localhost:9099'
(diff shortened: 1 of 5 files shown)


functions/src/api/utils/seeds/seeds.ts
```diff
@@ -0,0 +1,39 @@
+import { Request, Response } from 'express'
+import { wrap } from 'async-middleware'
+import { seedAuth } from './authSeeds'
+import { seedFirestore } from './seedFirestore'
+import { TEST_ACCOUNTS } from '../../../db/constants'
+
+export const seeds = wrap(async (req: Request, res: Response) => {
+    const authEmulatorUrl = 'localhost:9099'
+    const firestoreEmulatorUrl = 'localhost:9099'
```
MattiaPrimavera, line 9: isn't it strange the port is the same ?

MattiaPrimavera (approved):
Good job, I suggest to automate the process and run the seed script after the emulators are run

functions/src/api/utils/seeds/seeds.ts
HugoGresse: it is strange 🧐
but it works too...

PR #462, opened 2020-12-31
[FRONT] Fix newsletter modal opening from NavigationDrawer

ntchou:
- use redux to manage generic modal
- todo use it for the others modals
- moved Newsletter from pages to components
Peek 2020-12-31 01-10


src/components/Footer/Footer.tsx
```diff
@@ -56,9 +60,11 @@ const Footer = () => {
...
                 </li>
                 <li className="footer-link">
-                    <Link to="" className="footer-link-content">
-                        <NewsletterModal />
-                    </Link>
+                    <button
+                        className="footer-link-content"
+                        onClick={showNewsletterModal}>
+                        {t('footer.newsletter')}
+                    </button>
```
MattiaPrimavera, line 67: Footer is not visible anymore (removed). I think I had to left some files since it was affecting the UI when removing

src/components/NewsletterModal/NewsletterModal.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import Newsletter from './Newsletter'
+import Modal from '../Modal/Modal'
+import { ModalProps } from '../Modal/types'
+
+export default function NewsletterModal({ onClose, isOpened }: ModalProps) {
+    return (
+        <Modal isOpened={isOpened} onClose={onClose}>
+            <Newsletter />
```
MattiaPrimavera, line 9: may we also dinamically set the children to use, instead of creating a component for every modal ?

MattiaPrimavera (changes requested):
Nice job!!!

src/components/Footer/Footer.tsx
ntchou: indeed

src/components/NewsletterModal/NewsletterModal.tsx
ntchou: yes it we'll be great.

MattiaPrimavera (approved):

PR #519, opened 2021-01-03
[FRONT] fix edit room hide call confirmation block

ntchou:
- translation
- refactoring
Peek 2021-01-03 20-47

### DIFF
--- a/src/pages/AdminDashboard/CreateRoomConfirmation.tsx
+++ b/src/pages/AdminDashboard/CreateRoomConfirmation.tsx
@@ -1,6 +1,7 @@
 import React from 'react'
 import Typography from '@material-ui/core/Typography'
 import Paper from '@material-ui/core/Paper'
+import Box from '@material-ui/core/Box'
 import { useTranslation } from 'react-i18next'
 import { useSelector, useDispatch } from 'react-redux'
 import { selectCreatedRoom } from './roomsSelector'
@@ -8,7 +9,15 @@ import { Link } from 'react-router-dom'
 import { CopyToClipboard } from 'react-copy-to-clipboard'
 import { preventDefault } from './UserDetails'
 import { Button } from './CreateRoomForm/CreateRoomForm'
-import { newRoom } from './roomsActions'
+import { resetRoomCreated } from './roomsActions'
+import { makeStyles } from '@material-ui/core'
+
+const useStyles = makeStyles({
+    paper: {
+        padding: '1em',
+        border: '2px dashed',
+    },
+})
 
 export default function CreateRoomConfirmation() {
     const { t } = useTranslation('dashboard')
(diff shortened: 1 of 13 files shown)


src/pages/AdminDashboard/CreateRoomForm/CreateRoomForm.tsx
```diff
@@ -170,7 +170,8 @@ const CreateRoomForm = ({ fields, onFormSubmit, onCreateFormReset }) => {
         <Formik
             enableReinitialize
             initialValues={fields}
-            //validationSchema={validationSchema}
+            validationSchema={validationSchema}
+            validateOnBlur={false}
```
MattiaPrimavera, line 174: isn't it this option we discussed yesterday ?

src/pages/AdminDashboard/CreateRoomConfirmation.tsx
```diff
@@ -21,16 +30,29 @@ export default function CreateRoomConfirmation() {
...
         <>
             <Typography variant="h5" component="h1">
-                {t('confirmation.heading', { roomName: 'toto' })}
+                {t('confirmation.heading', { roomName: createdRoom.name })}
             </Typography>
-            <Paper>{t('confirmation.message', { roomName: 'toto' })}</Paper>
+            <Box m={2} />
+            <Paper elevation={0} className={classes.paper}>
+                {t('confirmation.message', {
+                    hostName: '', //todo not return yet by api,
```
MattiaPrimavera, line 47: I think I saw @HugoGresse fix on this one

MattiaPrimavera (changes requested):

src/pages/AdminDashboard/CreateRoomForm/CreateRoomForm.tsx
ntchou: Yes. I have started yup implementation. Only two fields are required a this Time.

src/pages/AdminDashboard/CreateRoomConfirmation.tsx
HugoGresse: fixed in https://github.com/Instant-Visio/InstantVisio-WebApp/pull/518

HugoGresse: API deployed, getRooms now return the hostName

HugoGresse: hmm, I'm not sure it solve this through

ntchou: i just saw it now. `hostName` is not saved in firestore. Is it planned. We need it when we edit a room

HugoGresse: hostName are saved in firestore, on creation or edit

ntchou: my bad

ntchou: is only add if we set `sendAt` (and if we have `destinations` and `participants`).  `sendAt` is for reminder?

HugoGresse: yes, `sendsAt` is to create reminder, alongside hostName and destinations. It will be better to createReminderthis way too to prevent sending invitation to destination. 
No sendsAt with destinations = invite participant directly
sendsAt with destinations = create reminder. 

i'll update createRoom to save hostName no matter the case though.

HugoGresse: sendsAt is an array of seconds timestamp too, so you can create multiple reminders at once.

MattiaPrimavera: There's an opened issue on that: #477

HugoGresse: API has been deployed and now always save the hostName

MattiaPrimavera (approved):

Issue #514, opened 2021-01-02
Join a group btn visible only on mobile builds


MattiaPrimavera:
- Currently the `join group` btn is also visible on mobile browsers
- fix `isAndroid` function to check if the app was built for android

<img width="309" alt="Screenshot 2021-01-03 at 13 29 41" src="https://user-images.githubusercontent.com/11927970/103478547-cabac300-4dc7-11eb-8467-f9610f8845de.png">

### Expected result
- the button should be visible on android application
- the button should not be visible on mobile browsers

PR #527, opened 2021-01-03
Join a group btn visible only on mobile builds #514

kangadominique123:
Capture d’écran 2021-01-03 à 15 55 54
Capture d’écran 2021-01-03 à 15 56 15
Capture d’écran 2021-01-03 à 16 48 53

show btn join-group only in **android-app** and **desktop-web**

### DIFF
--- a/src/services/platform.ts
+++ b/src/services/platform.ts
@@ -4,7 +4,10 @@ export const isIos = () => isPlatform('ios')
 
 export const isAndroid = () => isPlatform('android')
 
-export const isAndroidAndWeb = () => !getPlatforms().some(r=>["desktop","electron","pwa"].indexOf(r) >= 0);
+export const isNative = () =>
+    !getPlatforms().some(
+        (r) => ['desktop', 'electron', 'pwa', 'mobileweb'].indexOf(r) >= 0
+    )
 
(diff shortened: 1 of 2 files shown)


MattiaPrimavera:
The button should normally be visible only on mobile applications (even android only is fine for the moment), but not the on web apps

src/services/platform.ts
```diff
@@ -1,8 +1,10 @@
-import { isPlatform } from '@ionic/react'
+import { isPlatform, getPlatforms } from '@ionic/react'
 
 export const isIos = () => isPlatform('ios')
 
 export const isAndroid = () => isPlatform('android')
 
+export const isAndroidAndWeb = () => !getPlatforms().some(r=>["desktop","electron","pwa"].indexOf(r) >= 0);
```
MattiaPrimavera, line 7: Is the function body correctly returning if the platform is `android` ?
- The button should only be shown on `Android` built app (not browser)
- if you want the button can also be displayed on `iOs`
- no browser should display the button

MattiaPrimavera (changes requested):

MattiaPrimavera (approved):

Issue #537, opened 2021-01-04
update account quota on room creation


ntchou:
When we create a room with participants account quota is not refreshed

PR #538, opened 2021-01-04
[FRONT] #537 update quota displaying on room creation

ntchou:


src/pages/AdminDashboard/roomsActions.ts
```diff
@@ -68,6 +69,9 @@ export const createRoom = (t, room: Room, remindAt: number) => async (
         const { destinations } = room
 
         dispatch(setRoomCreated(roomId, countDestinations(destinations)))
+        if (destinations.length) {
+            dispatch(getUserDetails(t))
```
MattiaPrimavera, line 73: shouldn't we always get the rooms list, but only display the meeting still to happen ?

MattiaPrimavera (changes requested):

src/pages/AdminDashboard/roomsActions.ts
ntchou: this is only used to refresh quota/usage block (on the right in admin dashboard)

MattiaPrimavera (approved):

PR #554, opened 2021-01-06
Feature/547 autocomplete suggestions for group

HugoIneat:
- add getGroups + redux associated classes
- add autocomplete suggestions
- update validation to manage groups
- add error translations

### DIFF
--- a/src/pages/AdminDashboard/CreateRoomForm/createRoomTools.ts
+++ b/src/pages/AdminDashboard/CreateRoomForm/createRoomTools.ts
@@ -1,5 +1,6 @@
 import { parsePhoneNumber } from 'libphonenumber-js'
 import { UNITS } from '../Reminders/NotificationSelector'
+import { Group } from  '../groupsSelector'
 
 export const mapDestinationsToInputField = (destinations) => {
     const values = [] as any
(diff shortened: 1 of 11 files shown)


src/pages/AdminDashboard/CreateRoomForm/createRoomTools.ts
```diff
@@ -46,8 +47,13 @@ export const validatePhoneNumber = (phoneNumber) => {
     return isValid
 }
 
-export const isParticipantValid = (participant) => {
-    return validateEmail(participant) || validatePhoneNumber(participant)
+export const validateGroupName = (groupName : string, groups : Array<Group>): boolean => {
+    const groupFound = groups.find(group => group.name === groupName);
+    return groupFound !== undefined;
```
MattiaPrimavera, line 52: what about

```js
return group || false
```

HugoGresse (approved):

MattiaPrimavera (changes requested):

MattiaPrimavera:
There's a conflict to solve

src/pages/AdminDashboard/CreateRoomForm/createRoomTools.ts
HugoIneat: According to find documentation :

The value of the first element in the array that satisfies the provided testing function. Otherwise, undefined is returned.

MattiaPrimavera: Yes, but your validate function returns a `boolean`. I just found this way of expressing that more elegant than testing against `undefined` value. I think the result should be the same :)

MattiaPrimavera (approved):

PR #582, opened 2021-01-08
Enforce value set

HugoIneat:
Enforce value set for now/later toggle button

### DIFF
--- a/src/pages/AdminDashboard/CreateRoomForm/CreateRoomForm.tsx
+++ b/src/pages/AdminDashboard/CreateRoomForm/CreateRoomForm.tsx
@@ -77,7 +77,9 @@ const CreateRoomForm = ({ fields, onFormSubmit, onCreateFormReset }) => {
         event: React.MouseEvent<HTMLElement>,
         newValue: string | null
     ) => {
-        setToggleValue(newValue)
+        if(newValue !== null) {
+            setToggleValue(newValue)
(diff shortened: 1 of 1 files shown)


src/pages/AdminDashboard/CreateRoomForm/CreateRoomForm.tsx
```diff
@@ -77,7 +77,9 @@ const CreateRoomForm = ({ fields, onFormSubmit, onCreateFormReset }) => {
         event: React.MouseEvent<HTMLElement>,
         newValue: string | null
     ) => {
-        setToggleValue(newValue)
+        if(newValue !== null) {
+            setToggleValue(newValue)
```
MattiaPrimavera, line 81: why is this needed, considering the state has already an initial value ?

MattiaPrimavera (changes requested):

src/pages/AdminDashboard/CreateRoomForm/CreateRoomForm.tsx
HugoIneat: When you unselect a button, the value returned is null. In fact it disable the possibility to unselect a button.

ref : https://material-ui.com/components/toggle-button/#enforce-value-set

MattiaPrimavera (approved):

Issue #580, opened 2021-01-08
EditRoom handle reminders


MattiaPrimavera:
When editing a room:
- remove all reminders previously created
- create again all reminders sent by the front

PR #583, opened 2021-01-09
API: add reminders edit on editRoom, fix createGroup and add startAt on push data

HugoGresse:
- reminders can be edited on editRoom endpoint if sendsAt is supplied
- owner of a group is automatically added to the group member (so he can see those groups upon getting all his groups)
- startAt field to push data to we can adjust the notification sounds if the room does not start now
- check for unique participant name on all the room existing user no matter if they are there or not (issue happen during real testing)

Fix https://github.com/Instant-Visio/InstantVisio-WebApp/issues/580

### DIFF
--- a/functions/src/api/v1/rooms/editRoom.ts
+++ b/functions/src/api/v1/rooms/editRoom.ts
@@ -5,6 +5,9 @@ import { Timestamp } from '../../../firebase/firebase'
 import { RoomDao, RoomEditData } from '../../../db/RoomDao'
 import { BadRequestError } from '../../errors/HttpError'
 import { parseDestinations } from '../utils/parseDestinations'
+import { parseSendsAt } from './parseSendsAt'
+import { ReminderDao } from '../../../db/ReminderDao'
+import { RoomId } from '../../../types/Room'
 
 /**
  * @swagger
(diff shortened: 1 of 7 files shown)


HugoGresse:
Review ? 😅

functions/src/api/v1/rooms/editRoom.ts
```diff
@@ -72,5 +84,26 @@ export const editRoom = wrap(async (req: Request, res: Response) => {
...
+    if (sendsAtValues.length > 0) {
+        const oldReminders = await ReminderDao.listByRoomId(roomId)
+
+        for (const sendAt of sendsAtValues) {
+            const sendAtMillis = parseInt(sendAt) * 1000
+            const sendAtTimestamp = Timestamp.fromMillis(sendAtMillis)
+            await ReminderDao.add(roomId, sendAtTimestamp)
+        }
+
+        for (const reminder of oldReminders) {
```
ntchou, line 105: this is shouldn't be executed before add?

HugoGresse: the goal is to prevent removing any reminders if the add op fail, so it's kind like a transaction but ery very basic.

ntchou (approved):

PR #589, opened 2021-01-12
Feature/581 group members list

HugoIneat:
Add group members list component.


src/components/GroupMembersList/groupActions.ts
```diff
@@ -0,0 +1,54 @@
...
+    const group = selectGroup(getState())
+    const api = new Api(token)
+
+    if(!group?.members){
+        return
+    }
+
+    const newGroup = {
+        ...group,
+        members :  _.differenceBy(group.members, members,'id')
```
ntchou, line 42: why using `lodash`? You can use es6 native functions 
```
members.filter(new => !group.members.some(old => old.id === new.id))

```

src/components/GroupMembersList/groupActions.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    try {
+        const group = await api.getGroup(groupId)
+        dispatch(setGroup(group))
+    } catch (err) {
+        dispatch(showErrorMessage(t('errors.group-fetch')))
+    }
+}
+
+export const deleteMembers = (t, groupId, members) => async (dispatch, getState) => {
```
ntchou, line 31: better typing?

src/components/GroupMembersList/groupActions.ts
```diff
@@ -0,0 +1,54 @@
...
+
+
+export const setGroup = (group: any): GroupActionsTypes => ({
+    type: SET_GROUP,
+    payload: {
+        group,
+    },
+})
+
+export const getGroup = (t, groupId) => async (dispatch, getState) => {
```
ntchou, line 19: better typing?

src/components/GroupMembersList/GroupMembersList.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { selectGroup } from './groupSelector'
+import { useSelector } from 'react-redux'
+import { selectUser } from '../../components/App/userSelector'
+import { useTranslation } from 'react-i18next'
+import { getGroup, deleteMembers } from './groupActions'
+import { useDispatch } from 'react-redux'
+
+
+
+export const preventDefault = (event: React.SyntheticEvent) =>
```
ntchou, line 22: not used

src/components/GroupMembersList/groupActions.ts
```diff
@@ -0,0 +1,54 @@
...
+    if(!group?.members){
+        return
+    }
+
+    const newGroup = {
+        ...group,
+        members :  _.differenceBy(group.members, members,'id')
+    }
+    
+    dispatch(setGroup(newGroup))
```
ntchou, line 45: i think you should `dispatch` after `api delete`

HugoIneat: I decided to do it before api delete for a better UX experience. Here the user can click on the 'X', the member will be deleted form the group state and be removed from the list. Then the api is called and if it fails an error will be displayed and the state of the group will rollback. (and the list will be refreshed).

I you prefere the other way I can change it. (click 'X', show loader and wait for the api result to remove the member UI side)

ntchou: it's ok ;)

src/components/GroupMembersList/GroupMembersList.tsx
```diff
@@ -0,0 +1,91 @@
...
+                </Grid>
+                <Grid item>
+                    <Typography variant="h6">
+                        {t('title')}
+                    </Typography>
+                </Grid>
+            </Grid>
+            <Divider />
+            <List component="nav" className={classes.list}>
+                {group && group.members.map((member) => {
```
ntchou, line 73: It's just an advice. I know we used to do this but prefer ternary instead. If `group` return `0` you'll render `0`  the result will always be 0 because in javascript `0 && anything` because 0 is falsy, so it doesn't evaluate the right side of the `&&`

ntchou (approved):

PR #594, opened 2021-01-16
Add two channel and manage them for push notif

HugoGresse:
- add two new channel id (and change the android default one) in order to have one channel with sound and very important, the other less important without sound. 
- change notification push type if the room is starting in less than 5 min. If it's more, it a reminder push notif, otherweise it's a sound notif with "URGENCY". 
- manage channel id in functions

### DIFF
--- a/src/services/local-notifications.ts
+++ b/src/services/local-notifications.ts
@@ -5,7 +5,12 @@ import {
 } from '@capacitor/core'
 const { LocalNotifications } = Plugins
 export class LocalNotificationsService {
-    static async schedule(title?: string, body?: string, extra?: any) {
+    static async schedule(
+        title?: string,
+        body?: string,
+        extra?: any,
+        channelId: string = 'visio-call-notifications'
+    ) {
         const ONE_SECOND_FROM_NOW = new Date(Date.now() + 1000)
         const notifs = await LocalNotifications.schedule({
             notifications: [
@@ -18,7 +23,7 @@ export class LocalNotificationsService {
                     actionTypeId: '',
                     extra,
                     smallIcon: 'ic_logo_mobile',
-                    channelId: 'visio-call-notifications',
+                    channelId: channelId,
                 },
             ],
         })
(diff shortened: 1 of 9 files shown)


src/services/local-notifications.ts
```diff
@@ -5,7 +5,12 @@ import {
 } from '@capacitor/core'
 const { LocalNotifications } = Plugins
 export class LocalNotificationsService {
-    static async schedule(title?: string, body?: string, extra?: any) {
+    static async schedule(
+        title?: string,
+        body?: string,
+        extra?: any,
+        channelId: string = 'visio-call-notifications'
```
MattiaPrimavera, line 12: By default should not simulate a call in my advice

src/services/push-notifications.ts
```diff
@@ -13,18 +13,30 @@ const { PushNotifications } = Plugins
 export class PushNotificationsService {
     static createDefaultChannel() {
         const notificationChannel: NotificationChannel = {
-            id: 'visio-call-notifications', // id must match android/app/src/main/res/values/strings.xml's default_notification_channel_id
-            name: 'Visio call notifications',
-            description: 'Visio call notifications',
+            id: 'visio-call-notifications',
```
MattiaPrimavera, line 16: Same as above

MattiaPrimavera (changes requested):

src/services/push-notifications.ts
HugoGresse: I've changed the default_channel_id for the android manifest which remove any specifics of the default channel and is not used in the app

src/services/local-notifications.ts
```diff
@@ -18,7 +23,7 @@ export class LocalNotificationsService {
                     actionTypeId: '',
                     extra,
                     smallIcon: 'ic_logo_mobile',
-                    channelId: 'visio-call-notifications',
+                    channelId: channelId,
```
ntchou, line 26:
```suggestion
                    channelId,
```

src/services/local-notifications.ts
ntchou, line 12: Type is inferred. No need to precise

ntchou (approved):

